PR #2, opened 2020-02-11
Reiterated the quickstart documentation

niclashedam:
I have updated the quickstart a little to not litter in the user's home folder and remove information that I think is unimportant.


safl:
This is great, thanks!

PR #7, opened 2020-03-13
Reproduce

niclashedam:
This is a very early first iteration of the reproduce feature. After completing a test run you can call `cij_bundle` on the path, which will create a `.cijoe` file in the current working directory. Move this to the new environment and call `cij_reproduce` on the file. This will create a temporary path, where the test plan is rerun. After running it will print out the number of passed, failed and unknowns. When the metric parser is complete I suggest that those numbers will be shown in the print as well.

### DIFF
--- a/bin/cij_runner
+++ b/bin/cij_runner
@@ -6,6 +6,8 @@ import argparse
 import uuid
 import sys
 import os
+import glob
+import datetime
 import cij.runner
 import cij.util
 import cij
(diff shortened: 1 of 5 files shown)


niclashedam:
I have added a short description to the docs. Read that for reference.

niclashedam:
Rewritten as per out discussion. As the testplan argument is required, I have made an exception such that if you pass a directory as the testplan, it will be treated as a request to reproduce a given result.

safl:
Picked on ``rework`` branch with an added ``TODO`` note in commit-message.

PR #8, opened 2020-03-24
Initial Docker container

niclashedam:
The Dockerfile will create a docker container that is equivalent to running the _cijoe_ command. I have created a docker repo on [DockerHub](https://hub.docker.com/repository/docker/cijoe/cijoe). We should setup travis to push to there when releases are created.

### DIFF
--- a/bin/cijoe
+++ b/bin/cijoe
@@ -47,6 +47,14 @@ if [[ ! -f "$ENV_FPATH" ]]; then
 		exit
 	fi
 fi
+
+# check if we are in docker
+if [ -f /.dockerenv ]; then
+    cp -R /tmp/cijoe/.ssh /root/.ssh
+    chmod 700 /root/.ssh
+    chmod 600 /root/.ssh/*
+fi
+
 # shellcheck disable=SC2016
 bash --rcfile <(echo '[[ -f ~/.bashrc ]] && . ~/.bashrc; source modules/cijoe.sh; cd '"$HERE"'; export PS1="[cijoe:'"$ENV_NAME"'] $PS1"; source '"$ENV_FPATH"'; cij::emph "Welcome to CIJOE ENV_FPATH: '"$ENV_FPATH"'"; ')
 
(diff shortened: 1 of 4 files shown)


docs/Makefile
```diff
@@ -16,7 +16,7 @@ configure: clean
 
 .PHONY: kmdo
 kmdo:
-	kmdo src
+	kmdo src || exit 0
```
safl: Please split this into its own commit w.g. "docs:build: osx related work-around"

docs/src/docker.rst
```diff
@@ -0,0 +1,41 @@
...
+  docker run -it \
+  -v $(pwd)/target_env.sh:/cijoe/target_env.sh \
+  -v $(pwd)/results:/results \
+  cijoe
+
+You can now run your testplan. See `QuickStart`_ for examples on how to use
+**cijoe**. Starting the docker container is equivalent to running the **cijoe**
+command.
+
+.. note:: If you run **cijoe** on a remote system, bear in mind that you may
```
safl: Would be great with documentation on how to transfer keys and manage populate the ssh-agent inside the container.

docs/Makefile
niclashedam: Moved into a related OSX commit on #7.

docs/src/docker.rst
niclashedam: Alright. I have added a small feature that eases the pain of adding the keys to the docker container. It is not so easy to do for others due to the strict permission requirements. I have made the feature such that you can mount your SSH folder to `/tmp/cijoe/.ssh`, and then cijoe will do the rest.

I have also added a warning to the docs saying that this is insecure with untrused code, as the software has full access to the private key.

bin/cijoe
```diff
@@ -47,6 +47,14 @@ if [[ ! -f "$ENV_FPATH" ]]; then
 		exit
 	fi
 fi
+
+# check if we are in docker
+if [ -f /.dockerenv ]; then
```
safl: Please use the ``[[`` ``]`` bash-ism

safl: Nevermind, I'll pull it in and edit that.

PR #10, opened 2020-04-07
Docker fixes

niclashedam:
This PR contains two commits that fix some issues with the Docker.

1. Behind the scenes, the cijoe will check if the _.ssh_ folder is mounted to the container. If so, it will move the config file out to the home folder, if it exists, and change the permissions. It will also use the ssh-agent to import all keys such that there is no permission issues. This might be more uncessary work than before, but it makes the Docker container behave in a more logical and prediticable way.

2. I have removed the _mktemp_ alias. It is now up to the user to make sure to put results in the /results folder. The docs have been updated to reflect this along with an example of how to do it easily.


niclashedam:
By the way, we need to setup travis to push docker images to DockerHub on release.

safl:
Great thanks!

PR #14, opened 2021-01-28
modules: fix pylint errors

micvbang:


micvbang:
Sorry, didn't mean to post this yet

modules/cij/runner.py
```diff
@@ -340,6 +372,8 @@ def trun_to_file(trun, fpath=None):
 def trun_to_junitfile(trun, fpath=None):
     """Generate jUNIT XML from testrun YML"""
 
+    # pylint: disable=too-many-locals
+
```
safl, line 376: I guess the stamp code could be simplified and remove locals. E.g.
```
if trun.stamp:
  trun.stamp["begin"] = trun.stamp.get("begin", time.time())
  trun.stamp["end"] = trun.stamp.get("end", time.time())
```

safl (changes requested):
Great, a couple of minor thing inline.

modules/cij/runner.py
micvbang: Ah yeah, that's a great simplification!

I think putting the variables into a dict instead of using real variables is kind of worsening the code just to satisfy pylint warnings. In my opinion the code would be better if try to eliminate a variable elsewhere in the scope or keep the disable hint :)
I'll try to :dagger: a variable elsewhere!

modules/cij/runner.py
```diff
@@ -80,12 +96,14 @@ class TestCase(Runnable):
...
     @staticmethod
-    def tcases_from_dicts(tcase_dicts) -> list[TestCase]:
+    def tcases_from_dicts(tcase_dicts) -> List[TestCase]:
         """
         Deserialize a list of tcase dictionaries into a list of TestCases
         """
-        complex_keys = {'hooks': Hook.hooks_from_dict} 
-        tcase_fields = set(TestCase.__dataclass_fields__)
+        complex_keys = {'hooks': Hook.hooks_from_dict}
+
```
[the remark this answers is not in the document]
micvbang: Nice catch - that was not on purpose

modules/cij/runner.py
```diff
@@ -150,6 +176,12 @@ def tsuites_from_dicts(tsuites_dicts) -> list[TestSuite]:
 
 @dataclasses.dataclass
 class TestRun:
+    """
+    TestRuns contain information required to execute a testplan.
+    """
+
+    # pylint: disable=too-many-instance-attributes
```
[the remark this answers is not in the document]
micvbang: Yeah, it seems like it. My google-fu doesn't turn up any other solutions to this

modules/cij/runner.py
micvbang: I updated this and managed to simply even further. I think I'd personally keep the pylint-disable over what I came up with (because of readability), but I think both are okay

safl: But the stamp-field is already a dict, right? The proposal just does not use local-vars to give them a default-value?

modules/cij/runner.py
```diff
@@ -372,35 +366,24 @@ def trun_to_file(trun, fpath=None):
...
-            if not stamp_end:
-                stamp_end = time.time()
-
-            if stamp_end > stamp_begin:
-                duration = stamp_end - stamp_begin
+        stamp = trun.stamp
+        duration = max(
+            0,
+            stamp.get("end", time.time()) - stamp.get("begin", time.time())
+        )
```
micvbang, line 380: We can skip the if-statement altogether because the dataclass promises us that `trun.stamp` always contains a dict.

Another variation of the above is as follows, but requires the pylint-disable:

```python
        stamp_begin = trun.stamp.get("begin", time.time())
        stamp_end = trun.stamp.get("end", time.time())
        duration = max(0, stamp_end - stamp_begin) 
```

modules/cij/runner.py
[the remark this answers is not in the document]
safl: Yeah ok. Perhaps flake8 would do a better job here...

modules/cij/runner.py
safl: Ah, what you did here seems nice, what I proposed was wrong... since yeah... it should not alter the trun, just derive the duration.

modules/cij/runner.py
micvbang: Ahhh, you're completely right. I don't know why I read this like we were introducing `trun.stamp["stamp_begin"]` and `trun.stamp["stamp_end"]`

modules/cij/runner.py
safl: How about:
```
duration = max(
    0,
    trun.stamp.get("end", time.time()) - trun.tamp.get("begin", time.time())
)
```

Ain't that a winner?

safl: Or... just:
```
trun.stamp["begin"] = trun.stamp.get("begin", time.time())
trun.stamp["end"] = trun.stamp.get("end", time.time())
```
if it makes sense to update the trun...

modules/cij/runner.py
```diff
@@ -373,11 +374,9 @@ def trun_to_junitfile(trun, fpath=None):
...
         doc_testsuites = doc.createElement('testsuites')
 
-        stamp = trun.stamp
-        duration = max(
-            0,
-            stamp.get("end", time.time()) - stamp.get("begin", time.time())
-        )
+        stamp_begin = trun.stamp["begin"] or time.time()
+        stamp_end = trun.stamp["end"] or time.time()
+        duration = max(0, stamp_end - stamp_begin)
```
safl, line 379:
```
duration = max(0,
    stamp["end"] or time.time() -
    stamp["begin"] or time.time()
)
```

safl: How about this?

micvbang: Deal!

safl:
Great, then just squash the last one.

safl:
Thanks!

PR #15, opened 2021-01-29
Makeover flake8

safl:
Adding flake8 to self-test and fixed what it found except for three annotation issues.

### DIFF
--- a/.github/workflows/selftest.yml
+++ b/.github/workflows/selftest.yml
@@ -35,6 +35,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install graphviz wheel pylint
+        pip install graphviz wheel flake8
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
 
     - name: Build
--- a/.github/workflows/selftest.yml
+++ b/.github/workflows/selftest.yml
@@ -34,8 +34,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install graphviz wheel pylint
-        pip install graphviz wheel flake8
+        pip install graphviz wheel pylint flake8
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
 
     - name: Build
--- a/bin/cij_fetch
+++ b/bin/cij_fetch
@@ -19,6 +19,7 @@ except ImportError:
 
 CHUNK_SIZE = 16 * 1024
 
+
 def md5_create(path):
     """
     Create md5sum of <path> and store it in <path>.md5
@@ -36,6 +37,7 @@ def md5_create(path):
             md5_fd.write(out)
 
     return process.returncode
+
 
 def meta_local(path):
     """Equivalent of meta_reservoir but on local filesystem"""
@@ -60,6 +62,7 @@ def meta_local(path):
 
     return None
 
+
 def meta_reservoir(url):
     """Returns {"length": ..., "md5": ...} on url in reservoir"""
 
@@ -82,12 +85,14 @@ def meta_reservoir(url):
 
     return None
 
+
 def meta_is_equiv(one, other):
     """Return True two different metas has the same size and md5sum."""
 
     return (one is not None and other is not None) and \
-            (one["length"] == other["length"]) and \
-            (one["md5"] == other["md5"])
+        (one["length"] == other["length"]) and \
+        (one["md5"] == other["md5"])
+
 
 def reservoir_to_mem(url):
     """Returns reservoir resource from url."""
@@ -97,21 +102,22 @@ def reservoir_to_mem(url):
 
     return content
 
+
 def reservoir_to_file(url, path):
     """Downloads the url to path"""
 
     url_meta = meta_reservoir(url)
     local_meta = meta_local(path)
 
-    if meta_is_equiv(local_meta, url_meta): # Skip if it meta is equivalent
+    if meta_is_equiv(local_meta, url_meta):     # Skip if it meta is equivalent
         return True
 
-    local_dir = os.path.dirname(path)       # Ensure parent dir exists
+    local_dir = os.path.dirname(path)           # Ensure parent dir exists
     if not os.path.exists(local_dir):
         os.makedirs(local_dir)
 
     try:
-        response = urlopen(url)             # Download the file
+        response = urlopen(url)                 # Download the file
         with open(path, 'wb') as res_fd:
             while True:
                 chunk = response.read(CHUNK_SIZE)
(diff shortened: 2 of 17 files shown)


.github/workflows/selftest.yml
```diff
@@ -35,6 +35,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install graphviz wheel pylint
+        pip install graphviz wheel flake8
```
micvbang, line 38: I think it would make sense to move `flake8` up to the line above

testcases/cijoe_flake8.sh
```diff
@@ -0,0 +1,47 @@
...
+paths=""
+for path in {bin,hooks,testcases}/* "${CIJ_PKG_REPOS}/modules/cij/"*.py "${CIJ_PKG_REPOS}/setup.py"; do
+  if [[ -d "$path" ]]; then
+    continue
+  fi
+
+  if grep "^#.*python" "$path" > /dev/null || [[ $path == *.py ]]; then
+    paths="$paths $path"
+  fi
+done
```
micvbang, line 32: This is bash voodoo magic

modules/cij/__init__.py
```diff
@@ -110,13 +111,15 @@ def expand_path(path):
 
     conf = {v: os.environ.get("_".join([prefix, v])) for v in names}
 
-    for env in (e for e in conf.keys() if e[:len(prefix)] in names and conf[e]):
+    for env in (e for e in conf.keys()
+                if e[:len(prefix)] in names and conf[e]):
         conf[env] = expand_path(conf[env])
         if not os.path.exists(conf[env]):
```
micvbang, line 117: I think the readability of this filter would be improved if it was assigned to a variable and given a name

bin/cij_fetch
```diff
@@ -82,12 +85,14 @@ def meta_reservoir(url):
...
+
 def meta_is_equiv(one, other):
     """Return True two different metas has the same size and md5sum."""
 
     return (one is not None and other is not None) and \
-            (one["length"] == other["length"]) and \
-            (one["md5"] == other["md5"])
+        (one["length"] == other["length"]) and \
+        (one["md5"] == other["md5"])
+
```
micvbang, line 96: To long expressions like this one, I usually add a parenthesis around the whole thing. It allows one to skip the line continuations, e.g.

```
return (
  (one is not None and other is not None) and
  (one["length"] == other["length"]) and
  (one["md5"] == other["md5"])
)
```

modules/cij/runner.py
```diff
@@ -416,7 +430,8 @@ def trun_to_junitfile(trun, fpath=None):
                     doc_failure = doc.createElement("failure")
                     doc_failure.setAttribute(
                         "message",
-                        "not executed" if tcase.rcode is None else "test failed"
+                        "not executed" if tcase.rcode is None
+                        else "test failed"
```
micvbang, line 434: I would personally indent the `else` to align with the `if`

modules/cij/runner.py
```diff
@@ -576,10 +591,10 @@ def tsuite_setup(trun, declr, enum) -> TestSuite:
     tcase_fpaths = []                               # Load testcase fpaths
     if os.path.exists(suite.fpath):              # From suite-file
         suite_lines = (
-            l.strip() for l in open(suite.fpath).read().splitlines()
+            line.strip() for line in open(suite.fpath).read().splitlines()
```
micvbang, line 594: The opened file is never closed - consider doing something like

```
suite_content = ""
with open(suite.fpath) as suitef:
  suite_content = suitef.read()
  
suite_lines = ( .... )
  
```

bin/cij_tlint
```diff
@@ -60,8 +60,12 @@ def tsuitefile_testcases_exists(ident, roots, fnames):
 
     for ts_fname in fnames["TSUITES"]:    # Search .suite files
         ts_fpath = os.sep.join([roots["TSUITES"], ts_fname])
-        ts_lines_all = (l.strip() for l in open(ts_fpath).read().splitlines())
-        ts_lines = (l for l in ts_lines_all if len(l) > 1 and l[0] != "#")
+        ts_lines_all = (
+            line.strip() for line in open(ts_fpath).read().splitlines()
```
micvbang, line 64: Same as above -- this file is never closed

bin/cij_tlint
```diff
@@ -83,8 +87,12 @@ def testcase_unused(ident, roots, fnames):
 
     for ts_fname in fnames["TSUITES"]:  # Testcases from suite-files
         ts_fpath = os.sep.join([roots["TSUITES"], ts_fname])
-        ts_lines_all = (l.strip() for l in open(ts_fpath).read().splitlines())
-        ts_lines = (l for l in ts_lines_all if len(l) > 1 and l[0] != "#")
+        ts_lines_all = (
+            line.strip() for line in open(ts_fpath).read().splitlines()
```
micvbang, line 91: Same as above -- this file is never closed

micvbang (commented):
Looks good - here are some inline comments

modules/cij/runner.py
safl: Good point, thanks

modules/cij/runner.py
safl: I am going with:
```
 "not executed" if tcase.rcode is None else
 "test failed"
```
Linters to not appreciate the long indent to match up to ``if``

bin/cij_fetch
safl: Much nicer, thanks.

modules/cij/__init__.py
safl: Agree, I've adjusted accrodingly.

.github/workflows/selftest.yml
safl: Agree, fixed.

testcases/cijoe_flake8.sh
safl: Yeah, no way around that :)

safl:
Update with changes according to feedback.

modules/cij/runner.py
```diff
@@ -430,8 +430,8 @@ def trun_to_junitfile(trun, fpath=None):
                     doc_failure = doc.createElement("failure")
                     doc_failure.setAttribute(
                         "message",
-                        "not executed" if tcase.rcode is None
-                        else "test failed"
+                        "not executed" if tcase.rcode is None else
+                        "test failed"
```
micvbang, line 434: Ah sorry, I wasn't clear enough on this one; I meant

```python
"not executed" if tcase.rcode is None
                        else "test failed"
```
but the version you just posted actually works quite well!

micvbang:
LGTM

PR #16, opened 2021-01-29
Micvbang/cij analyser

micvbang:
PR for discussion on direction

### DIFF
--- a/bin/cij_extractor
+++ b/bin/cij_extractor
@@ -1,0 +1,82 @@
+#!/usr/bin/env python
+"""
+ CIJOE test extractor
+"""
+from __future__ import print_function
+import argparse
+import sys
+import os
+
+import cij.extractor
+import cij.runner
+import cij.util
+import cij
+
+
+def parse_args():
+    """Parse command-line arguments for cij_extractor"""
+
+    cij_evars = cij.paths_from_env("CIJ", ["TEMPLATES"])
+    for key, val in cij_evars.items():
+        if val is None:
+            cij.err(
+                f"rprtr:rprtr: failed parsing environment variable 'CIJ_{key}'"
+            )
+            return None
+
+    # Parse the Command-Line
+    prsr = argparse.ArgumentParser(
+        description='cij_extractor - CIJOE Test data extractor',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
+    prsr.add_argument(
+        '--output',
+        help="Path to test output",
+        default=os.getcwd()
+    )
+    prsr.register('action', 'custom_extend', cij.util.ExtendAction)
+    prsr.add_argument(
+        "--extractor",
+        help="Name of, or path to, one or more extractor(s)",
+        action="custom_extend",
+        choices=[e.name for e in cij.extractor.find_extractors()],
+        nargs="+",
+        required=True,
+        type=str
+    )
+
+    args = prsr.parse_args()
+
+    args.output = cij.util.expand_path(args.output)
+    if not os.path.exists(args.output):
+        cij.err(f"rprtr:output: {args.output}, does not exist")
+        return None
+
+    args.trun_fpath = cij.runner.yml_fpath(args.output)
+    if not os.path.exists(args.trun_fpath):
+        cij.err(f"rprtr:trun_fpath: {args.trun_fpath}")
+        return None
+
+    return args
+
+
+def main():
+    """
+    Parse environment variables and command-line arguments constructing a
+    configuration for which to invoke the extractor
+    """
+
+    args = parse_args()
+    if args is None:
+        cij.err("rprtr: failed parsing command-line args")
+        return 1
+
+    rcode = cij.extractor.main(args)
+    if rcode:
+        cij.err(f"rprtr: rcode: {rcode}, error while analysing {args.output}")
+
+    return rcode
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/modules/cij/analyser.py
+++ b/modules/cij/analyser.py
@@ -1,0 +1,321 @@
+#!/usr/bin/env python3
+"""
+    Library functions for cij_analyser
+"""
+
+import re
+import os
+from typing import List, Tuple
+import dataclasses
+import copy
+
+import yaml
+
+from cij.runner import Status, TestRun, TestCase
+import cij.runner
+import cij
+from cij.util import rehome
+from cij.errors import CIJError, InvalidRangeError, UnknownUnitError
+
+UNITS = {
+    # general
+    '': 1,          # no unit
+    'B': 1,         # bytes
+    'k': 1000,      # kilo
+    'M': 1000**2,   # mega
+    'G': 1000**3,   # giga
+
+    # kibi
+    'KiB': 1024**1,  # kibibytes
+    'MiB': 1024**2,  # mibibytes
+    'GiB': 1024**3,  # gibibytes
+    'TiB': 1024**4,  # tibibytes
+
+    # kilo
+    'kB': 1000**1,  # kilobytes
+    'MB': 1000**2,  # megabytes
+    'GB': 1000**3,  # gigabytes
+    'TB': 1000**4,  # gigabytes
+
+    # time
+    'nsec': 1/1000**3,  # nanoseconds
+    'usec': 1/1000**2,  # microseconds
+    'msec': 1/1000**1,  # milliseconds
+    'sec': 1,         # seconds
+    'min': 60       # minutes
+}
+
+
+class Range:
+    """
+    Range implements parsing and validation of mathematical range notation,
+    e.g. `[-5;100[` which translates to "must be >= -5 and < 100".
+    """
+
+    # pylint: disable=no-self-use
+    # pylint: disable=too-few-public-methods
+
+    _rng_re = re.compile(
+        r"^(?P<elower>\[|\])\s*(?P<rstart>-inf|-?\d+(\.\d*)?)\s*;"  # [1.0;
+        r"\s*(?P<rend>inf|-?\d+(\.\d*)?)\s*(?P<eupper>\[|\])"       # 1.0]
+        fr"\s*(?P<unit>({'|'.join(UNITS)}))$"                      # ms
+    )
+
+    def __init__(self, rng: str):
+        match = self._rng_re.match(rng)
+        if not match:
+            raise InvalidRangeError(f"invalid range \"{rng}\"")
+
+        rng_start = float(match["rstart"])
+        rng_end = float(match["rend"])
+        if rng_start > rng_end:
+            raise InvalidRangeError(
+                "expected lower bound <= upper bound, "
+                f"{rng_start} <= {rng_end}"
+            )
+
+        if match["unit"] not in UNITS:
+            raise UnknownUnitError(
+                f"unexpected unit '{match['unit']}', "
+                f"supported units are: {list(UNITS.keys())}"
+            )
+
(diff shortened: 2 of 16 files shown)


safl:
This is really shaping up.
A thought, how about adding the unittests as part of the ``cij`` package and make then runnable via module load e.g. ``python  -m cij.selftest", then a cijoe-testcase can easily be added running the Python unittest.

micvbang:
Yeah, that's an excellent idea!

modules/cij/__init__.py
```diff
@@ -52,43 +52,46 @@ def get_time_stamp():
     return ""
 
 
-def info(txt):
+def info(txt, fd=sys.stdout):
```
safl, line 55: Put these changes to the output functions into their own commit describing what they enable

micvbang: Roger!

safl:
Can you rebase on latest and squash linter-fixes?
Also, for "cij: add output file to log functions", this is great, but nothing is using that behavior, right?
I guess it should be possible to either provide a flag telling cijoe to dump the output log inside the output-directory, or another full-path argument to allow the outmessaging to go into a completely separate location.
Or how about, "--log-file" taking a path, and if it is a filename-only then place it relative to results output.

micvbang:
See inline answers

> Can you rebase on latest and squash linter-fixes?

Yep!

> Also, for "cij: add output file to log functions", this is great, but nothing is using that behavior, right?
Yeah, I'm using it [here](https://github.com/refenv/cijoe/pull/16/files#diff-be0e4a41bcac7c70ae06680b13bfb22aca09d5a42d5c7d3960f859ece497a092R176) to format the log-output generated by the analyser, which the reporter reads and includes in the runlogs [here](https://github.com/refenv/cijoe/pull/16/files#diff-769ccd09ca0b6da17cd20e96d9cd11ad003d362d5900caa6429a4a4b0eb1f4d9R92)

> I guess it should be possible to either provide a flag telling cijoe to dump the output log inside the output-directory, or another full-path argument to allow the outmessaging to go into a completely separate location.
> Or how about, "--log-file" taking a path, and if it is a filename-only then place it relative to results output.

micvbang:
Depending on whether we do or don't like there to be coupling between the analyser and the reporter (as there is in this PR right now) I guess we could instead add another field in the trun, e.g. `trun.testcases[].tcase[].extra_runlogs`, to tell the reporter to include these files. That would remove the need for the reporter to explicitly know about the plog format

safl:
> Depending on whether we do or don't like there to be coupling between the analyser and the reporter (as there is in this PR right now) I guess we could instead add another field in the trun, e.g. `trun.testcases[].tcase[].extra_runlogs`, to tell the reporter to include these files in the runlog for that tcase. That would remove the need for the reporter to explicitly know about plog files and allow us to put other stuff in the runlogs without having to change the reporter again in the future

It would be preferable that the postprocessing stuff simply produces its output-files in the _aux folder

micvbang:
Alright, makes sense. I think it would be nice to coordinate the last steps of this on a call :)

safl:
Let's do that :)

modules/cij/__init__.py
```diff
@@ -81,6 +81,7 @@ def err(txt, file=sys.stdout):
     print("%s# %s%s%s" % (PR_ERR_CC, get_time_stamp(), txt, PR_NC), file=file)
     _flush(file)
 
+
```
safl, line 84: This is probably nit-picking, but these should go into the first commit as that is where the spaces where removed in the first-place :)

modules/cij/__init__.py
```diff
@@ -91,6 +92,7 @@ def emph(txt, rval=None, file=sys.stdout):
     else:               # any other value, considered 'bad'
         err(txt, file=file)
 
+
```
safl, line 95: Same here

modules/cij/analyser.py
```diff
@@ -16,6 +16,34 @@
...
 
+_units = {
+    # general
+    '': 1,          # no unit
+    'B': 1,         # bytes
+    'k': 1000,      # kilo
+    'M': 1000**2,   # mega
+    'G': 1000**3,   # giga
+
+    # kibi
```
safl, line 27: This is awesome. But I think that it should use the IEC prefixes for the 1024^n and the SI for the 1000^n as described here: https://en.wikipedia.org/wiki/Binary_prefix

modules/cij/analyser.py
```diff
@@ -16,6 +16,34 @@
...
+    'KB': 1024**1,  # kibibytes
+    'MB': 1024**2,  # mibibytes
+    'GB': 1024**3,  # gibibytes
+    'TB': 1024**4,  # tibibytes
+
+    # kilo
+    'kB': 1000**1,  # kilobytes
+    'mB': 1000**2,  # megabytes
```
safl, line 35: MB

modules/cij/analyser.py
```diff
@@ -16,6 +16,34 @@
...
+_units = {
+    # general
+    '': 1,          # no unit
+    'B': 1,         # bytes
+    'k': 1000,      # kilo
+    'M': 1000**2,   # mega
+    'G': 1000**3,   # giga
+
+    # kibi
+    'KB': 1024**1,  # kibibytes
```
safl, line 28: KiB, MiB, GiB, TiB

modules/cij/analyser.py
micvbang: Good point - this is fixed now

safl:
For the commit message: ``mod:cij:reporter: extract rehome to util `` please change it to:

``Fix rehome() for dataclasses and move to cij.util``

modules/cij/extractor.py
```diff
@@ -0,0 +1,86 @@
...
+        return 1
+
+    try:
+        extractor = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(extractor)
+    except FileNotFoundError:
+        cij.err(f"Cannot find: {name}")
+        raise InitializationError(f"extractor '{name}' not recognized")
+
+    if not getattr(extractor, 'extract_metrics', None):
```
safl, line 62: Great.

modules/cij/extractors/__init__.py
```diff
@@ -0,0 +1 @@
+from cij.extractors import fio_json_read
```
safl, line 1: Why is this needed?

modules/cij/analyser.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+def preqs_from_file(fpath):
+    """ Read yaml-formatted performance requirements from fpath """
+
+    with open(fpath, 'r') as preqf:
+        return yaml.safe_load(preqf)
+
+
+def plog_from_file(fpath):
```
safl, line 143: The "plog" is deprecated now, right?

safl: Or is that the "analysis" log?

testrequirements/example_06_prequirements.preq
```diff
@@ -0,0 +1,6 @@
+global:
```
safl, line 1: Does it support this:

```
global:
  iops: "[0;500]k"
  bwps: "[80;200]MiB"
  lat:  "[20;50]usec" 

testcases:
  xnvme_fioe:
    iops: "[0;800]k"
```

safl: That is, the global will be the default to compare all testcases to, then when "testcases" is provided then for specific testcases the metrics are overwritten.

safl (commented):
Looking great, a couple of questions :)

modules/cij/analyser.py
micvbang: Yep, good catch!

modules/cij/extractors/__init__.py
micvbang: Hmm, that's a good question. I'm using it to make `fio_json_read` available as an import from `cij.extractors`. It's quite possible that we can remove this, but then I think we'll have to change the way that imports of the extractors work.

It would be quite nice to not have to add an import for each extractor we add - it's very easy to forget

testrequirements/example_06_prequirements.preq
micvbang: Not yet - forgot to look at that after our chat yesterday. But it's definitely a nice feature. It shouldn't be difficult to add

modules/cij/extractors/__init__.py
safl: Yeah, the not so pretty thing to do would be in ``__init__`` doing:
``from cij.extractors import *``

Some PEP will probably complain about that :)

testrequirements/example_06_prequirements.preq
safl: Would be great!

micvbang: Fixed

modules/cij/extractors/fio_json_read.py
```diff
@@ -0,0 +1,55 @@
...
+                pmetric,
+                extr_name=_MY_NAME,
+                fname=os.path.basename(fpath),
+                job_id=n,
+            )
+
+            metrics.append({
+                "ctx": ctx,
+                "iops": job["read"]["iops_mean"],
+                "bwps": job["read"]["bw_mean"] * 1024,
```
safl, line 39: Perhaps this should be some fancy function utilizing the units?

safl: E.g. calling some metric-function saying "I provide **x** in unit **y** give me **x** in default-unit"

micvbang: That would be awesome. Should I give an implementation a try?

I'm thinking something like what we discussed the other day: `to_base_unit(bwps, 'KiB')` which would then convert `bwps` to bytes (since that's out base unit for storage)

micvbang: Hehe, you got here before me. Sounds like we're thinking the same thing!

safl: haha awesome, ``to_base_unit(bwps, 'KiB')`` sounds great!

modules/cij/extractors/__init__.py
micvbang: This should be fixed now - changed the way I'm listing extractors

modules/cij/extractors/fio_json_read.py
micvbang: This should be done now

safl: cool

modules/cij/extractors/fio_json.py
```diff
@@ -31,6 +32,7 @@ def make_context(fio_obj: dict, fname: str, job_id: int) -> dict:
         "iodepth": opt["iodepth"],
         "fname": fname,
         "job_id": job_id,
+        "extractor_name": extr_name,
```
safl, line 35: great!

bin/cij_extractor
```diff
@@ -0,0 +1,82 @@
...
+        'output',
+        help="Path to test output",
+        default=os.getcwd()
+    )
+    prsr.register('action', 'custom_extend', cij.util.ExtendAction)
+    prsr.add_argument(
+        "--extractor",
+        help="Name of, or path to, one or more extractor(s)",
+        action="custom_extend",
+        choices=[e.name for e in cij.extractor.find_extractors()],
```
safl, line 42: Great!

micvbang: Thanks! I randomly stumbled upon it and found that it fit very well. It even lists the existing ones if you don't supply one

safl:
Thanks!

PR #20, opened 2021-02-03
Makeover fixes

safl:
A fix to the ``ssh:cmd`` and addition an addition of ``cij::cmd``.
The point of ``cij::cmd`` is to provide a helper in testcases for the actual command-invocation, such that one does write ``ssh::cmd`` directly in the testcase but rather via ``cij::cmd`` instead, ``cij::cmd`` then invokes the command via a transport such as SSH or via local command execution. It is in preparation for fixing up that lack on indirection but added now such that new examples can use it.

### DIFF
--- a/modules/cij.sh
+++ b/modules/cij.sh
@@ -193,3 +193,63 @@ cij::isint() {
   [[ "$1" =~ ^[0-9]+$ ]]
   return $?
 }
+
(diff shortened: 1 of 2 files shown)


micvbang:
LGTM - I like this abstraction

PR #21, opened 2021-02-03
Makeover fio

safl:
This is a example of wrapping fio with a cijoe testcase that can be instrumented from the testplan.
It also serves as an example of the type of data produced by a testcase which needs further post-processing in the form of extraction from output logs, transformation into formats for plotting and performance metric evaluation.

### DIFF
--- a/testcases/extc_05_fio.sh
+++ b/testcases/extc_05_fio.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+#
+# Run fio controlled by environment variables, formal and json-formatted output stored in AUX
+#
+# Instrument the fio-wrapper script via the environment variables:
+#
+# FIO_BIN: set this to the fio binary to use. Def: "fio"
+# FIO_ARGS: Assign this for auxilary arguments to fio. Def: ""
+#
+# FIO_NRUNS: Repeat the fio invocation this many times. Def: "1"
+# FIO_BS_LIST: The "--bs" to run fio for, provide a string on the form "512 4k 32k". Def: "512"
+# FIO_IODEPTH_LIST: The "--iodepth" to use; provide a string on the form "1 2 4 8". Def: "1"
+#
+# FIO_FILENAME: When set, adds argument "--filename=${FIO_FILENAME}"
+# FIO_SCRIPT: When set, adds argument "--script=${FIO_SCRIPT}"
+# FIO_SECTION: When set, adds argument "--section=${FIO_SCRIPT}"
+#
+# shellcheck disable=SC2119
+#
+CIJ_TEST_NAME=$(basename "${BASH_SOURCE[0]}")
+export CIJ_TEST_NAME
+# shellcheck source=modules/cijoe.sh
+source "$CIJ_ROOT/modules/cijoe.sh"
+test::enter
+
+: "${FIO_BIN:=fio}"
+
+: "${FIO_ARGS:=}"
+
+: "${FIO_NRUNS:=1}"
+: "${FIO_BS_LIST:=512}"
+: "${FIO_IODEPTH_LIST:=1}"
+
+main() {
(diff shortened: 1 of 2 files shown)


testcases/extc_05_fio.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+#
+# Run fio controlled by environment variables
+#
+# Requires that:
+#
```
micvbang, line 6: Requirements are missing

testcases/extc_05_fio.sh
```diff
@@ -0,0 +1,68 @@
...
+source "$CIJ_ROOT/modules/cijoe.sh"
+test::enter
+
+: "${FIO_BIN:=fio}"
+
+: "${FIO_ARGS:=}"
+
+: "${FIO_NRUNS:=1}"
+: "${FIO_BS_LIST:=512}"
+: "${FIO_IODEPTH_LIST:=1}"
```
micvbang, line 21: Does this evaluate to a list? I don't think I understand the semantics of `for FIO_IODEPTH in $FIO_IODEPTH_LIST; do` in l.26

safl: They assign the variable when it is unset or null. Basically just providing a default-value.

testcases/extc_05_fio.sh
```diff
@@ -0,0 +1,68 @@
...
+          return 1
+        fi
+
+        # Download the output
+        if ! cij::pull "${_target_fpath}" "${_output_fpath}"; then
+          cij::err "xnvme::fioe: failed pulling down fio output-file"
+          return 1
+        fi
+
+        # Dump output files it to stdout for prosperity
```
micvbang, line 55: I think maybe this should have been "posterity"? Not sure :thinking:

testcases/extc_05_fio.sh
safl: Updated with description of usage.

testplans/example_05_fio.plan
```diff
@@ -0,0 +1,42 @@
...
+  The testcase requires that certain variables such as FIO_BIN is set.
+
+  Depending on what and how fio is used then it makes sense to either define it in the
+  test-environment, or in the testplan. Example
+
+  * We want to utilize fio to measure bandwidth of a device, then simply set the FIO_BIN in the
+    target-environment as it is a precondition for what we want to do
+  * We want to measure differences in two different fio version on the same device, then it would
+    make sense to define paths to the different versions in the plan per at the testsuite level.
+    And utilize the alias feature to tell the two apart easility
```
micvbang, line 12: easility -> easily

testcases/extc_05_fio.sh
safl: Absolutely true :)

micvbang (approved):
Looks good! Only have minor text suggestions

testcases/extc_05_fio.sh
safl: Changed it to describe the actual reason.

testplans/example_05_fio.plan
safl: Fixed :)

safl:
Fixed based on feedback and rebased.

testcases/extc_05_fio.sh
micvbang: Ah ok, cool. But is the default value a list? Or am I misunderstanding the semantics of the `for xx in xx_list; do` construct? I feel like it should iterate over a list, but the default value being `1` doesn't seem listy to me

micvbang: Ah okay, I think I understand it now.. Found a bash example that described it

micvbang:
LGTM

PR #23, opened 2021-02-08
Support for multiple testplans per testrun

safl:
The basic functionality is in place; the runner can execute multiple testplans and visualize multiple plans in a single report. However, the following needs attention before merge:

* ~~The report-template, it is currently showing testplan info such as evars, hook-output etc. as part of the testrun, this should be moved down into the testplan-section.~~
* ~~The states of: status, progress, and wallc needs some verification, they are not accounted correctly.~~
  * ~~TestPlan now has "progress", however, it is not accounted correctly, this needs fixing.~~

Providing it now for early review and to determine possible collisions with the analyse feature.

### DIFF
--- a/modules/cij/runner.py
+++ b/modules/cij/runner.py
@@ -793,7 +793,7 @@ def main(conf):
 
             for tcase in (tc for tc in tsuite.testcases if tsuite.entered):
                 tcase_match = conf.get("TESTCASE_MATCH", None)
-                if not (tcase_match and tcase_match not in tcase.name):
+                if tcase_match is None or tcase_match in tcase.name:
                     tcase.enter(trun)
                     if tcase.entered:
                         script_run(trun, tcase)
--- a/modules/cij/runner.py
+++ b/modules/cij/runner.py
@@ -40,15 +40,88 @@ class Status:
 @dataclasses.dataclass
 class Runnable:
     """
-    Runnable contains attributes required for classes to be run by script_run
+    Runnable contains the common attributes of testplans, testsuites,
+    testcases, and hooks as these are in different way "run".
     """
+    # pylint: disable=too-many-instance-attributes
+    # There are alot of attributes, which for this class is fine.
+
+    entered: bool = False
+
+    ident: str = "ANON"
     name: str = "UNNAMED"
-    evars: dict = dataclasses.field(default_factory=dict)
-    res_root: str = ""              # Abs. path to result directory
+
+    fpath_orig: str = ""            # Abs. path to script / suite /  plan
     fpath: str = ""                 # Abs. path to script / suite /  plan
+    fname: str = ""                 # Filename without path
+
+    evars: dict = dataclasses.field(default_factory=dict)
+    hooks: dict = dataclasses.field(default_factory=lambda: {
+        "enter": [],
+        "exit": []
+    })
+    hnames: list = dataclasses.field(default_factory=list)
+
+    res_root: str = ""              # Abs. path to result directory
+    aux_root: str = ""              # Abs. path to auxiliary directory
+    aux_list: list = dataclasses.field(default_factory=list)
+
     log_fpath: str = ""             # Abs. path to run.log
+
+    status: str = Status.Unkn
     rcode: Optional[int] = None
+    stamp: dict = dataclasses.field(default_factory=lambda: {
+        "begin": None,
+        "end": None
+    })
     wallc: Optional[float] = None
+
+    def enter(self, trun):
+        """Called by runner before invoking the runnable"""
+
+        if trun.conf["VERBOSE"]:
+            cij.emph("rnr:enter { ident: %r }" % self.ident)
+
+        self.stamp["begin"] = time.time()
+
+        rcode = 0
+        for hook in self.hooks["enter"]:  # ENTER-hooks
+            rcode = script_run(trun, hook)
+            if rcode:
+                break
+
+        self.entered = not rcode
+
+        if trun.conf["VERBOSE"]:
+            cij.emph("rnr:enter { ident: %r, rcode: %r } " % (
+                self.ident, rcode
+            ))
+
+        return rcode
+
+    def exit(self, trun):
+        """Called by runner after invoking the runnable"""
+
+        if trun.conf["VERBOSE"]:
+            cij.emph("rnr:exit: { ident: %r }" % self.ident)
+
+        rcode = 0
+        if self.entered:
+            for hook in reversed(self.hooks["exit"]):      # EXIT-hooks
+                rcode = script_run(trun, hook)
+                if rcode:
+                    break
+
+        self.stamp["end"] = time.time()
+        if self.wallc is None:
+            self.wallc = self.stamp["end"] - self.stamp["begin"]
+
+        if trun.conf["VERBOSE"]:
+            cij.emph("rnr:exit: { ident: %r, rcode: %r }" % (
+                self.ident, rcode
+            ))
+
+        return 0
 
 
 @dataclasses.dataclass
@@ -57,8 +130,6 @@ class Hook(Runnable):
     Hooks are user-defined scripts that can be run before and/or after
     TestRuns, TestSuites, and TestCases.
     """
-    fname: str = ""
-    fpath_orig: str = ""
 
     @staticmethod
     def hooks_from_dict(hooks_dict) -> dict:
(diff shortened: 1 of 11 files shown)


templates/report.html
```diff
@@ -276,8 +276,16 @@ <h1 class="display-4">Results</h1>
   </div>
 </div>
 
+{% for tplan in dset.testplans %}
+
+<div class="jumbotron jumbotron-fluid">
+  <div class="container">
+    <h2 class="display-8">Testplan {{ tplan.name }} ( {{ tplan.ident }} )</h1>
```
halfzebra, line 283: Could be a typo here? The closing tag is different from the opening :thinking:

modules/cij/runner.py
```diff
@@ -181,6 +181,69 @@ def tsuites_from_dicts(tsuites_dicts) -> List[TestSuite]:
...
+
+        fields = _get_dataclass_fields(TestPlan) & set(tplan_dict)
+        for key in fields:
+            deserialize = complex_keys.get(key, lambda x: x)
+            setattr(tplan, key, deserialize(tplan_dict[key]))
+
+        return tplan
+
+    @staticmethod
+    def from_lod(lod) -> List[TestPlan]:
```
micvbang, line 243: `lod` means list of dicts, right? It took me a second to realize -- maybe it would be easier if we added a type annotation, e.g. `lod: Dict[str, Any]` (or a more specific type instead of Any, if that's possible)

modules/cij/runner.py
```diff
@@ -572,7 +630,7 @@ def tsuite_enter(trun: TestRun, tsuite: TestSuite) -> int:
     return rcode
 
 
-def tsuite_setup(trun: TestRun, declr, enum) -> TestSuite:
+def tsuite_setup(trun: TestRun, tplan, declr, enum) -> TestSuite:
```
micvbang, line 637: Since we're already type hinting most of our own classes, I think we should add it to `tplan: TestPlan` here as well

modules/cij/runner.py
```diff
@@ -588,10 +646,10 @@ def tsuite_setup(trun: TestRun, declr, enum) -> TestSuite:
     suite.alias = declr.get("alias")
     suite.ident = "%s_%d" % (suite.name, enum)
 
-    suite.res_root = os.sep.join([trun.conf["OUTPUT"], suite.ident])
+    suite.res_root = os.path.join(tplan.res_root, suite.ident)
     suite.aux_root = os.sep.join([suite.res_root, "_aux"])
```
micvbang, line 654: I think the change to `path.join` is nice - would make sense to do for aux_root as well. It guess it might make sense to just change all of these throughout the codebase in a separate commit

modules/cij/runner.py
```diff
@@ -681,80 +739,128 @@ def tcase_enter(trun: TestRun, tsuite: TestSuite, tcase: TestCase) -> int:
...
+    return tplan
+
+
+def trun_exit(trun: TestRun):
+    """Triggers when exiting the given testrun"""
+
+    if trun.conf["VERBOSE"]:
+        cij.emph("rnr:trun:exit")
+
+    return 0
```
micvbang, line 836: Should `trun.stamp["end"]` be written here?

modules/cij/runner.py
```diff
@@ -769,50 +875,62 @@ def main(conf):
...
+            ts_ent_err = tsuite_enter(trun, tsuite)
 
-        ts_err = 0
-        ts_ent_err = tsuite_enter(trun, tsuite)
-        for tcase in (tc for tc in tsuite.testcases if not ts_ent_err):
+            for tcase in (tc for tc in tsuite.testcases if not ts_ent_err):
+                tc_err = 0
 
-            tc_err = 0
+                tcase_match = conf.get("TESTCASE_MATCH", None)
```
micvbang, line 902: `tcase_match` assignment can be moved to before the `tplan` loop

modules/cij/runner.py
```diff
@@ -769,50 +875,62 @@ def main(conf):
...
+                if not (tcase_match and tcase_match not in tcase.name):
```
micvbang, line 903: I find double negations to be a bit difficult to grok - might be easier to understand if written `if not tcase_match or tcase_match in tcase.name`

templates/report.html
micvbang: Good catch!

modules/cij/runner.py
micvbang: This name sticks out from the naming of the other methods that does the same on our other classes. It might make sense to change all of them to this shorter naming `from_lod`, otherwise I think this one should be renamed `tplans_from_dicts`

templates/report.html
safl: That is a good catch! Thanks!

safl: fixed :)

micvbang (commented):
Looks good! I find it nice that the changes are mostly just adding the class and following the existing structure. Great!

I've got a few comments to changes made here, and then a few changes to code that wasn't changed in this commit. Sorry! Feel free to skip these for some other time.

At a higher level, I think this feature is super great. I think it will merge well with the analyser code, although the extra level of indirection (trun -> tplan) will make the extraction-part of the analyser slightly more work and a bit more complex to implement

modules/cij/runner.py
safl: I guess the type: ``lod: List[dict]`` would fit?
Regarding the naming, yeah, I'll switch to match what is there. Then a cleanup can be done later for changing naming overall.

micvbang: Ah yeah, sorry, I completely messed that type signature up!
I think we might get a linter error if we don't define what's in the dict, but give it a try

safl: mypy does not catch violations regardless of the annotation :/ Seem like it is because the function is applied through the "complex-key"  indirection, possibly just too much runtime polymorphism for mypy :)

modules/cij/runner.py
safl: Adding this to a list of cleanups to do later on.

modules/cij/runner.py
safl: fixed.

modules/cij/runner.py
safl: Time-stamping can't be done in ``on_exit``, since ``on_exit`` might not be invoked in case ``on_enter`` failed. Time-stamping is going on the list of things to improve... so does enter/exit, the logic in the main-runner loop is so convoluted due to this... the enter/exit functions *should* always be invoked, doing things like time-stamping, but hen just not execute hooks in case of enter-failure.

modules/cij/runner.py
safl: true, but it is unrelated to the multi-testrun, adding to list of things to cleanup.

modules/cij/runner.py
safl: Agree, putting it on the list of cleanups.

micvbang:
LGTM

modules/cij/runner.py
```diff
@@ -886,63 +820,59 @@ def main(conf):
...
-        for tsuite in (ts for ts in tplan.testsuites if not tp_ent_err):
-            ts_err = 0
-            ts_ent_err = tsuite_enter(trun, tsuite)
+    for tplan in (tp for tp in trun.testplans if trun.entered):
+        tplan.enter(trun)
 
-            for tcase in (tc for tc in tsuite.testcases if not ts_ent_err):
-                tc_err = 0
+        for tsuite in (ts for ts in tplan.testsuites if tplan.entered):
+            tsuite.enter(trun)
```
micvbang: We could avoid evaluating the static values `trun.entered`, `tplan.entered`, and `tsuite.entered` for every item in each list if we did something like

```python
if tplan.enter(trun):
  break
  
  for tsuite in tplan.testsuites:
    if tsuite.enter(trun):
      break
      
    for tcase in tsuite.testcases:
      ...
```
I think that this reads better for the three nested loops. But I can't come up with a good solution to skip the outer loop in the same way without wrapping the outermost loop in `if trun.entered:`, which would nest this thing even deeper

micvbang:
Nice, I like the cleanup of duplicate enter/exit code!

modules/cij/runner.py
safl: Yeah, there are also the other side-effects at the end of the loop-nests to update run exit() and assign status.
A way to make this prettier would probably be to move the "run" into the runnable and call recursively.

safl:
> Nice, I like the cleanup of duplicate enter/exit code!

Thanks, I like that a lot as well :) Thanks to that, then the addition of multiple testplans is only 30 lines of code in the runner-module. And there is still room to clean up additional attributes.

PR #24, opened 2021-02-10
Makeover conf

safl:
With this, then the Python utilities does not rely on having started the interactive cijoe environment.
That is, having invoked:

```bash
cijoe
# [2021-02-11 11:06:28] Welcome to CIJOE ENV_FPATH: /home/safl/.local/share/cijoe/envs/default.sh
[cijoe:default] safl@themachine:~$ 
```

The CIJOE Bash environment is great for doing things interactively, however, it should not be **required** when invoking the test-runner, reporter, etc. although should still utilize it when it is. So, ``cij.conf.Config`` is introduced, producing the installation paths based on the environment, when defined, and based on the result of ``cij_root`` when it is not.

The tools have been updated with the use of the class, and serialization/deserialization added.


modules/cij/runner.py
```diff
@@ -820,15 +820,17 @@ def main(conf):
...
+            trun_to_junitfile(trun)                 # Persist as jUNIT XML
 
         if tplan.exit(trun) or not tplan.entered:
             trun.status = tplan.status = Status.Fail
         if tplan.status == Status.Unkn and tplan.entered:
             tplan.status = Status.Pass
 
+        trun_to_file(trun)                      # Persist trun
+        trun_to_junitfile(trun)                 # Persist as jUNIT XML
+
```
micvbang, line 865: Nice little quality of life upgrade!

bin/cij_testcases
```diff
@@ -11,10 +11,11 @@ import os
 import dataclasses
 from cij.reporter import tcase_parse_descr, dset_to_html
 from cij.runner import TestCase
+import cij.conf
 import cij
 
 
-def construct_dset(evars):
+def construct_dset(conf):
```
micvbang, line 18: Would be nice to add type annotation

bin/cij_testcases
```diff
@@ -87,17 +83,17 @@ def parse_args():
     args.tmpl_fname = os.path.basename(args.tmpl_fpath)
     args.tmpl_name = os.path.splitext(args.tmpl_fname)[0]
 
-    return args, evars
+    return args, conf
 
 
-def main(args, evars):
+def main(args, conf):
```
micvbang, line 89: Would be nice to add type annotation for `conf`

modules/cij/runner.py
```diff
@@ -286,7 +288,12 @@ class TestRun(Runnable):
     # pylint: disable=too-many-instance-attributes
 
     ver: str = ""
-    conf: dict = dataclasses.field(default_factory=dict)
+    args: argparse.Namespace = dataclasses.field(
+        default_factory=argparse.Namespace
+    )
+    conf: cij.conf.Config = dataclasses.field(
```
micvbang, line 294: These type annotations are great -- we should do the same for some of the other fields, e.g. testcases and testsuites. I'll put that on my TODO

modules/cij/runner.py
```diff
@@ -311,8 +318,18 @@ def from_dict(trun_dict) -> TestRun:
         """
         Deserialize a trun dict into TestRun
         """
+
+        def args_from_dict(adict: dict) -> argparse.Namespace:
+            """Construct argparse.Namespace from dict"""
+
+            namespace = argparse.Namespace()
+            namespace.__dict__.update(adict)
```
micvbang, line 326: Neat

modules/cij/runner.py
```diff
@@ -467,10 +484,17 @@ def hooks_setup(trun: TestRun, parent, hnames=None) -> Dict[str, List[Hook]]:
 def trun_to_file(trun: TestRun, fpath=None):
     """Dump the given trun to file"""
 
+    def dict_factory(instance):
+        """huha"""
+
+        return dict(
+            (x[0], vars(x[1])) if x[0] == "args" else x for x in instance
+        )
```
micvbang, line 492: I think we can give a more precise docstring than this one :D

modules/cij/runner.py
```diff
@@ -732,7 +763,7 @@ def tplan_setup(trun: TestRun, tplan_fpath) -> TestPlan:
     return tplan
 
 
-def trun_setup(conf) -> TestRun:
+def trun_setup(args, conf) -> TestRun:
```
micvbang, line 766: Also type annotation for `conf` here

modules/cij/runner.py
```diff
@@ -790,8 +822,8 @@ def main(conf):
             tsuite.enter(trun)
 
             for tcase in (tc for tc in tsuite.testcases if tsuite.entered):
-                tcase_match = conf.get("TESTCASE_MATCH", None)
-                if tcase_match is None or tcase_match in tcase.name:
+                if args.testcase_match is None or \
+                        args.testcase_match in tcase.name:
```
micvbang, line 826: Could be written using parenthesis instead of the line continuation `\`

bin/cij_tlint
```diff
@@ -276,9 +277,9 @@ def main(roots):
 
 if __name__ == "__main__":
 
-    CIJ_ROOT = os.environ.get("CIJ_ROOT")
-    if not (CIJ_ROOT and os.path.exists(CIJ_ROOT)):
-        print("Please set ENV. VAR. CIJ_ROOT, correctly")
+    CONF = cij.conf.from_system()
+    if CONF is None:
+        cij.err("rprt: failed retrieving CIJOE configuration")
```
micvbang, line 282: Is there a reason this is called `CONF` here? I think it's called `conf` in other places

micvbang (commented):
Looks great! This is a very nice overall quality of life improvement in the usage of the tools.
I mostly got minor suggestions and comments on syntax stuff

micvbang:
Also some lovely commit messages btw -- I'll try to learn from them

bin/cij_tlint
safl: Yeah, it is in global-scope, some PEP says that variables in global-scope must be UPPERCASE.

modules/cij/runner.py
safl: Nice! Fixed.

modules/cij/runner.py
safl: Good idea. Adding for args as well.

modules/cij/runner.py
safl: Hah, true. Fixed.

bin/cij_testcases
safl: Good idea. Fixed.

bin/cij_tlint
micvbang: Ah, I didn't realize that -- cool!

safl:
Update based on review.

micvbang:
Great - LGTM

PR #25, opened 2021-02-15
Makeover plotter

safl:
A collection of fixes and then the addition of an initial metric-plotter implementation.

### DIFF
--- a/bin/cij_plotter
+++ b/bin/cij_plotter
@@ -1,0 +1,258 @@
+#!/usr/bin/env python
+"""
+ CIJOE Metric Plotter
+"""
+import argparse
+import hashlib
+import pprint
+import copy
+import sys
+import os
+import matplotlib.ticker as mticker
+import matplotlib.pyplot as plt
+import yaml
+import cij.analyser
+import cij.util
+import cij
+
+HATCHES = ['o', 'O', '.', '\\', '//', '-', '+', 'x', '\\', '*']
+
+MARKERS = ["+", "o", "s", "x", "h", "d", "*", "<", ">", "1", "*", "p"]
+
+METRICS = ["iops", "bwps", "lat"]
+
+METRIC_DEFAULT_UNIT = {
+    "iops": 'k',
+    "bwps": "MiB",
+    "lat": "usec"
+}
+
+KEY_DESCR = {
+    "iops": 'I/O operations Per Second',
+    "bwps": "BandWidth Per Second",
+    "lat": "Latency",
+
+    "bs": "Block-Size",
+    "iodepth": "I/O-Depth"
+}
+
+
+def collect_metrics(args):
+    """Collect all metrics found in 'metrics.yml' of testcase _aux folders"""
+
+    metrics = []
+    for root, _, _ in os.walk(args.output):
+        yml_fpath = os.path.join(root, "_aux", "metrics.yml")
+        if not (root.endswith(".sh") and os.path.exists(yml_fpath)):
+            continue
+
+        with open(yml_fpath, 'r') as yml_file:
+            metrics.extend(yaml.safe_load(yml_file))
+
+    return metrics
+
+
+def metrics_to_pdata(args, metrics):
+    """Transform the given metrics into plot-data"""
+
+    pdata = {}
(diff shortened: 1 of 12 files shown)


modules/cij/runner.py
```diff
@@ -448,36 +448,41 @@ def hook_setup(parent, hook_fpath) -> Hook:
...
 
+    """
     hooks: Dict[str, List[Hook]] = {
         "enter": [],
         "exit": []
     }
 
-    if hnames is None:       # Nothing to do, just return the struct
-        return hooks
+    if hnames is None:          # Nothing to do, just return the struct
```
micvbang, line 464: We no longer return the struct here

modules/cij/extractors/fio_json_read.py
```diff
@@ -33,6 +33,7 @@ def extract_metrics(tcase: TestCase) -> List[dict]:
                 fname=os.path.basename(fpath),
                 job_id=n,
             )
+            ctx["evars"] = tcase.evars
```
micvbang, line 36: Nice addition! I think it makes sense to add an `evar` arg to `fio_json.make_context`, so that we ensure that future extractors also set this value

bin/cij_plotter
```diff
@@ -0,0 +1,257 @@
+#!/usr/bin/env python
```
micvbang, line 1: Is it on purpose that this is implemented in a single file, unlike the runner/reporter/analyser?

micvbang, line 1: Is it on purpose that this is implemented in a single file, unlike the runner/reporter/analyser?

bin/cij_plotter
```diff
@@ -0,0 +1,257 @@
...
+    pdata = {}
+    for metric in metrics:                  # Transform metrics to plot-data
+        ctx = copy.deepcopy(metric["ctx"])
+
+        yval = metric[args.metric]          # metric-key to use as y-value
+
+        xval = int(ctx.pop(args.ctx))       # ctx-key to extract as x-value
+
+        if ctx["bs"] == "4k":               # ctx key/value requirements
+            continue
```
micvbang, line 67: Why do we not want to plot block size 4k?

micvbang (commented):
Looks very good! I have some minor things

bin/cij_plotter
micvbang: We could make this not depend on `ctx["bs"]` existing, e.g. `if ctx.get("bs", None) == "4k":

bin/cij_plotter
```diff
@@ -0,0 +1,257 @@
...
+
+    return metrics
+
+
+def metrics_to_pdata(args, metrics):
+    """Transform the given metrics into plot-data"""
+
+    pdata = {}
+    for metric in metrics:                  # Transform metrics to plot-data
+        ctx = copy.deepcopy(metric["ctx"])
```
micvbang, line 60: We could also make this independent on `metrix["ctx"]` by doing `metric.get("ctx", {})`

modules/cij/extractors/fio_json_read.py
safl: Agree, however, I could not quite see how to add it within the scope of the function... can you do a fix for it?

modules/cij/runner.py
safl: Ah, nice catch, fixing...

bin/cij_plotter
safl: Yeah, I just wanted something simple and self-contained.
In a second iteration, it should be part of the module code, where the plotter/data-set transformation are dynamically loadable from packages/module similar to the loading of extractors.

bin/cij_plotter
safl: Ah, nice catch!

bin/cij_plotter
safl: Oh, that is much better! Fixing...

modules/cij/extractors/fio_json_read.py
micvbang: Yep!

bin/cij_plotter
micvbang: Sounds good

PR #33, opened 2021-07-02
ssh: Do ssh copy id for user

Joelgranados:
One of the things that I want CIJOE to do for me is to setup the unprompted ssh connection.
This change activates the ssh-copy-id when the SSH_SETUP_UNPROMPTED is set.

One thing that I do not like is that when the ssh-copy-id fails because login is disabled on the remote host, I cannot tell that to the user when they are executing. 

Can we have  a way to communicate with the user?

### DIFF
--- a/modules/ssh.sh
+++ b/modules/ssh.sh
@@ -7,6 +7,7 @@
 # ssh::env       - Sets default vars for ssh wrapping
 # ssh::cmd <CMD> - Execute <CMD> using optional "SSH_CMD_TIMEOUT"
 # ssh::shell     - Get the regular shell using current environment
+# ssh::setup_unprompted() - Setup unprompted ssh connection to host
 #
 # REQUIRED variables:
 #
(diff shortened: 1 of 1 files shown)


safl:
To me this looks more like target provisioning/setup rather than actual utilization of the target. How about adding/refining the documentation on SSH:

- https://cijoe.readthedocs.io/en/latest/ssh.html
- https://github.com/refenv/cijoe/blob/master/docs/src/ssh.rst

Instead?

If you want to verify the SSH-setup described in a given target-environment, then you can verify it interactively by running:

```bash
cijoe your_target_env.sh
ssh::shell
```

Then instead of `ssh::setup_unprompted` a wrapper around `ssh-copy-id` could be added, assisting with the setup like:

```bash
cijoe your_target_env.sh
ssh::copy_id
ssh::shell
```

Joelgranados:
* Updating the docs is a good idea but it is not enough. The problem is that my run does not work and I don't really know its ssh so even if the docs are updated, ppl wont know to go there.

* Putting the copy_id exec in the ssh verification is a good idea as well. but this suffers from the same problem as before. I don't know that I need to run this verification when I hit a problem.

* If copy_id is not executed would you at least consider telling the user when the ssh authentication fails and how exactly to fix it. At the point of failure, we have all the information to exectute copy_id, we can just say. "ssh authentication error, please fix by funning this `ssh-copy-id ......`. And then the user does not have to hunt for this. they can just copy paste and execute.

PR #34, opened 2021-07-15
mod:py:runner: print error message with log path when testcase fails

micvbang:


micvbang:
I think it would make sense to lock the linter versions so that it doesn't become a moving goalpost.
Having people submit PRs and "being required" to fix linter errors of stuff they never touched is a bit daunting, I think

safl:
Thanks for the improvement and all these fixed.
I see your point. It is not very motivating to send a PR when it implies fixing unrelated stuff.
But thanks for doing so!

However, all those fixes are actually rather nice, perhaps, instead of fixing the linter version, then setup GHA to run daily, thus grabbing the latest linters and telling that there and things to improve.

micvbang:
Yeah, doing it automatically sounds great!

safl:
Thanks!

PR #35, opened 2021-07-23
Print the actual exception message wih FileExistsError is caught in c…

Joelgranados:
At least another type of FileExistsError can occur when catching this exception.
* the makedirs in runner.py(792) can fail because that directory already exists.

But then the user incorrectly thinks its the lock file cannot fix the error. 
I fixed it by actually outputing the caught error.

### DIFF
--- a/modules/cij/runner.py
+++ b/modules/cij/runner.py
@@ -14,7 +14,7 @@ import os
 import yaml
 import cij.conf
 import cij
-from cij.errors import CIJError, InitializationError
+from cij.errors import CIJError, InitializationError, InvalidResultAuxPath
 # pylint:disable=unsubscriptable-object
 
 HOOK_PATTERNS = {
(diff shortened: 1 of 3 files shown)


modules/cij/runner.py
```diff
@@ -789,7 +789,11 @@ def trun_setup(args: argparse.Namespace, conf: cij.conf.Config) -> TestRun:
     trun.res_root = args.output
     trun.aux_root = os.path.join(trun.res_root, "_aux")
 
-    os.makedirs(trun.aux_root)
+    try:
+        os.makedirs(trun.aux_root)
+    except CIJError as ex:
```
safl, line 794: Ah, now I get what you mean.
I don't think ``os.makedirs()`` would ever throw ``CIJError``?

How about catching ``FileExistsError`` here, then wrapping it in a new cijoe-exception e.g. ``InvalidResultAuxPath`` by adding it to ``modules/cij/errors.py`` and raising it with something like:
```
except FileExistsError as exc:
   cij.err("trun_setup: FAILED: os.makedirs(). Error: %r" % exc)
   raise InvalidResultAuxPath("foo") from exc
```

Joelgranados: I have updated the PR.

modules/cij/runner.py
```diff
@@ -814,6 +818,7 @@ def main(args, conf):
         trun = trun_setup(args, conf)   # Construct 'trun' from args and conf
     except CIJError as ex:
         cij.err("main:FAILED to start testrun: %s" % ex)
+        return 1
```
Joelgranados, line 821: I have added this return value when we catch an exception.
It is strange to me that it was not there before. Was it intended that this exception was ignored just with a print?

Joelgranados: I think this line returning is the one causing the self tests to fail
Still don't understand why we would silently ignore it.
One way I can fix is to catch the new InvalidResultAuxPath exception and return 1 and leave the CIJError as the back stop that will silently ignore the error.
What do you think?

Joelgranados: Think the auto-test are happy now.

safl:
closing as cijoe has been re-implemented.

PR #36, opened 2021-08-19
cij: Add cij::cmd_output to be able to capture the output

Joelgranados:
Just add a cmd that will not echo any additional output


safl:
This looks great. Can you add a sign-off to the commit?

PR #38, opened 2021-12-08
cij:extractors: search for extractors in cijoe_extractors

micvbang:
Refactor `cij_extractor` to find extractors from new `cijoe_extractors` namespace. This PR is dependent on the two new packages `cijoe-pkg-fio` and `cijoe-pkg-rocksdb` which implement this namespace using implicit namespace packages (PEP420)


safl:
There are a couple of linter-issues which looks like they are introduced here.
Regarding the "This PR is dependent on the two new packages cijoe-pkg-fio and cijoe-pkg-rocksdb which implement this namespace using implicit namespace packages (PEP420)", isn't that the other way around? the those two packages depend on cijoe providing the feature?

micvbang:
> There are a couple of linter-issues which looks like they are introduced here. 

Fixed!

>Regarding the "This PR is dependent on the two new packages cijoe-pkg-fio and cijoe-pkg-rocksdb which implement this namespace using implicit namespace packages (PEP420)", isn't that the other way around? the those two packages depend on cijoe providing the feature?

Hmm.. I would say that the dependency is circular. The packages _can_ be used without cijoe, but aren't really meant to. So they're kinda depending on cijoe. The extractor-functionality of cijoe does depend on packages being available in the `cijoe_extractors` package, so there's a dependency in that direction as well

micvbang:
Awesome!

PR #40, opened 2022-03-01
Non-zero exit on error when failing to setup testrun

halfzebra:
An exception to setup `trun` causes another exception later when we attempt to use the variable and leaves a lock file.

### DIFF
--- a/modules/cij/runner.py
+++ b/modules/cij/runner.py
@@ -814,6 +814,7 @@ def main(args, conf):
         trun = trun_setup(args, conf)   # Construct 'trun' from args and conf
     except CIJError as ex:
         cij.err("main:FAILED to start testrun: %s" % ex)
+        return 1
 
     trun_to_file(trun)              # Persist trun
     trun_to_junitfile(trun)         # Persist as jUNIT XML


safl:
In what way is behavior changed when doing this? Might it make sense to persist the ``trun`` before returning?

halfzebra:
This error actually happens when we try to persis the `trun`, here's the stack-trace:

```
# rnr: { OUTPUT: '/tmp/trun-268e7f70' }
# rnr:hooks_setup:FAIL { hname: 'sysinf' has no files }
# main:FAILED to start testrun: rnr:hooks_setup:FAIL { hname: 'sysinf' has no files }
Traceback (most recent call last):
  File "/home/smth/.local/bin/cij_runner", line 114, in <module>
    RES = main(ARGS)
  File "/home/smth/.local/bin/cij_runner", line 98, in main
    rcode = cij.runner.main(args, conf)
  File "/home/smth/.local/lib/python3.8/site-packages/cij/runner.py", line 818, in main
    trun_to_file(trun)              # Persist trun
UnboundLocalError: local variable 'trun' referenced before assignment
```

safl:
Ah I see, makes sense.
Can you adjust the commit:
* sign off the commit
* prefix the first line with ``runner: ...``
* add a description with the info you provided here, something like the below reformulated a bit:
```
An exception to setup trun causes another exception later when we attempt to use the variable and leaves a lock file.
We exit early if trun setup threw an exception preventing possible exceptions down the lline.
```

halfzebra:
I've replaced the `rnr` prefix with `runner` for consistency. Let me know if you want to go back to the short version.

safl:
Oh, I meant your commit-message. No need to refactor the "log-output-label", unless it is completely wrong.

safl:
Just keep the ``rnr`` prefix, otherwise you have the selftest failing as the line gets too long.

safl:
Excellent thank you! Now just a minor thing with that first commit-line, see:
https://github.com/refenv/cijoe/pull/40/commits/cf70534c5ea65aec6fa5c3cc74a875b3df0ad85a
It is sort or long, just make it super brief describing what it does, then provide the long description in "body" of the commit-message.

halfzebra:
Thanks for the feedback!

I have updated the commit with a shorter message and moved the description into the body :+1:

safl:
Thanks!

PR #47, opened 2022-09-23
setup: add cffi to the list of dependencies

Panky-codes:
Cijoe results in an error if cffi is not present. Add it to the list of deps while installing.

Error output:
    from . import _bcrypt
ModuleNotFoundError: No module named '_cffi_backend'

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,7 @@ setup(
         "scp",
         "setuptools>=60",
         "watchdog",
+        "cffi",
     ],
     python_requires=">=3.9",
     entry_points={


safl (commented):
When does this happen? I am just curious as I have not seen it.

Panky-codes:
> When does this happen? I am just curious as I have not seen it.

Just running `cijoe` gave this error in my setup, so I had to manually install cffi.

safl:
Curious, since ``cffi`` is a dependency of a dependency. @jwdevantier @baekalfen do you guys have a recommendation with such an issue? Is it a solution to add ``cffi`` as a requirement or what would you propose?

Baekalfen:
I would just add it. Unless there is a good reason for it, I would point the finger at `bcrypt` and say they didn't declare all their dependencies correctly. `bcrypt` is based on Rust, so it's not surprising they need `cffi`.

Baekalfen:
To rule out any issues; @Panky-codes should do a `pip install --upgrade pip` and make sure when you install `bcrypt` that it is a `...manylinux_2_28_x86_64.whl` package it downloads.

safl:
> 

Thanks for sharing the solution!

PR #50, opened 2023-01-30
processing: decode with errors='replace' when reading

karlowich:
Currently cijoe will run into errors like:
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xef in position 60: invalid continuation byte

This change fixes these errors


safl (approved):
Thanks!

PR #51, opened 2023-02-01
processing: fix bug introduced in PR#50

karlowich:
The decoding was implemented incorrectly, this corrects the error


safl:
Fingers crossed ;)

PR #52, opened 2023-02-03
Add description option for a workflow run and a clean up to --archive option

Panky-codes:
It would be useful to give a tag for a workflow run
while running cijoe during development. This option enables
to give a tag that will be stored in the workflow.state. The same
text will be used as a prefix with timestamp while storing the results
in the archive folder.
    
Example usage:
```bash
$ cijoe -c default.config -w example.workflow -t "directIO" --tag "io_uring"
$ ls -la cijoe-archive/
directIO-io_uring-2023-02-07_06:44:34
```

### DIFF
--- a/src/cijoe/cli/cli.py
+++ b/src/cijoe/cli/cli.py
@@ -75,9 +75,11 @@ def cli_archive(args):
     """Move 'output' directory into archive"""
 
     if args.output.exists():
-        archive = args.output.with_name("cijoe-archive") / str(
-            time.strftime("%Y-%m-%d_%H:%M:%S")
-        )
+        state_path = args.output / "workflow.state"
+        state = dict_from_yamlfile(state_path)
+        tag = "" if state["tag"] is None else '-'.join(state["tag"]) + '-'
+        t = str(time.strftime("%Y-%m-%d_%H:%M:%S"))
+        archive = args.output.with_name("cijoe-archive") / f"{tag}{t}"
         os.makedirs(archive)
         log.info(f"moving existing output-directory({args.output}) to '{archive}'")
         os.rename(args.output, archive)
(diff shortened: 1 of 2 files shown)


safl:
A couple of thoughts, how about:

* Calling this ``label`` instead of ``description``
* Supporting multiple labels when cli is given ``--label  foo --label bar --label baz``
  * First label is utilized for naming the archive
  * All labels are storing in the workflow-file

Panky-codes:
> A couple of thoughts, how about:
> 
> * Calling this `label` instead of `description`
> * Supporting multiple labels when cli is given `--label  foo --label bar --label baz`
>   
Done

>   * First label is utilized for naming the archive

Instead of using just the first label, I concatenate all labels while naming the folder in the archive as I feel that makes more sense, and avoids confusion for a new user.

safl:
How about calling this ``--tag``, ``-t`` instead? To avoid confusions/collision with ``--log-level / -l``?

safl:
Another thing, perhaps drop the removal of ``--archive``. The option is not superfluous, as what it provides is a means to archive an existing output-directory and do nothing else, e.g. calling ``cijoe --archive`` will "archive" an existing ``output`` folder and then exit.

Panky-codes:
> Another thing, perhaps drop the removal of `--archive`. The option is not superfluous, as what it provides is a means to archive an existing output-directory and do nothing else, e.g. calling `cijoe --archive` will "archive" an existing `output` folder and then exit.

Ahh. Got it! The same logic is open-coded in both places. Perhaps, I can change it to call the same function in both places

src/cijoe/cli/cli.py
```diff
@@ -190,13 +190,7 @@ def cli_workflow(args):
...
         return errno.EINVAL
 
-    if args.output.exists():
-        archive = args.output.with_name("cijoe-archive") / str(
-            time.strftime("%Y-%m-%d_%H:%M:%S")
-        )
-        os.makedirs(archive)
-        log.info(f"moving existing output-directory to {archive}")
-        os.rename(args.output, archive)
+    cli_archive(args)
```
safl, line 195: Nice catch! Thanks!

safl (approved):
This is great, thank you!

PR #58, opened 2024-05-07
fix(transport): add support for env variables in csh, cmd and pwsh

NaddiNadja:
The commands for setting environment variables in csh (default on FreeBSD) and in cmd and PowerShell (Windows) are different than in sh.

Which shell is used, should be set in the config file. If none is set, sh is assumed.

### DIFF
--- a/src/cijoe/core/transport.py
+++ b/src/cijoe/core/transport.py
@@ -86,6 +86,7 @@ class SSH(Transport):
         """Initialize the CIJOE SSH Transport"""
 
         self.config = config
+        self.shell = self.config.options.get("run").get("shell", "sh")
         self.output_path = output_path
 
         self.ssh = paramiko.SSHClient()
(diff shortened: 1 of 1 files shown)


safl (commented):
This is excellent @NaddiNadja , however, one thing I think we should consider is whether this should be part of different config-item. this could be "run" similar to workflows:

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrunshell

But without the "default". Something like:

```yaml
run:
  shell: "bash"
```

This would indicate that this modifies the behavior ``cijoe.run()``

Then, we could add something like:

```yaml
run:
  shell: "bash"
  env: {FOO: 123, BAR: 321}
```

As a place to agnostically define environemtn variables.

NaddiNadja:
@safl , I have added support for defining environment variables in the configuration file in the "run" item. If an environment variable is set in both the configuration file and in the parameters of the the `run()` call, it will use the variable from the parameters. 

So a config file with
```
run:
  env: { test: "some variable" }
```
And a workflow running
```python
cijoe.run('echo "$test"', env={"test": "another variable"})
```
Will produce the output
```
another variable
```

karlowich (approved):
Looks good

src/cijoe/core/transport.py
```diff
@@ -86,7 +86,7 @@ def __init__(self, config, output_path):
         """Initialize the CIJOE SSH Transport"""
 
         self.config = config
-        self.shell = self.config.options.get("run").get("shell", "sh")
+        self.shell = self.config.options.get("run", {}).get("shell", "sh")
```
safl, line 89: So, I know I initially mused on adding a top-level ``run`` item. That that I have seen this and mulled a bit on it, then it would be great if this was "top-level" under a ``cijoe`` item. Then, going forward we can eventually move the "transport" section into this as well.
By doing so, then the config-items for cijoe is the same as for everything else. So, for now we would have:

```yaml
cijoe:
  run:
    shell: bash
  env:
    foo: "this is so 42"
```

safl (changes requested):
This is excellent, I have a nit-pick and then a request for change as the work you have done here got me thinking about how configuration of cijoe itself should be managed.

src/cijoe/core/transport.py
NaddiNadja: This has been fixed now :)

safl (approved):
This is excellent, thank you @NaddiNadja

PR #63, opened 2024-06-04
feat(cijoe): merge `cijoe-pkg-*` into `cijoe`

NaddiNadja:
Closes #61

PR #64, opened 2024-06-04
refactor(config): move transport item to cijoe

### LINKED ISSUE #60: Refactor configuration
Currently all configuration-file items have a top-level identifier e.g. ``fio``, ``qemu``, etc. This clashes a bit with the configuration-items for **cijoe** itself, e.g. the configuration of cijoe-transports has top-level item ``transport``. For example, for the configuration-content:

```toml
[transport.ssh]
username = "root"
password = "root"
hostname = "localhost"
port = 4200
```

This could be under top-level item ``cijoe``. e.g.:

```toml
[cijoe.transport.ssh]
username = "root"
password = "root"
hostname = "localhost"
port = 4200
```

This will align nicely with the recent addition of specifying ``env`` and ``shell`` for ``cijoe.run``, which looks like:

```toml
[cijoe.run]
shell = "csh"
env = {FOO = "bar", JAZZ= "tazz"}
```

NaddiNadja:
Move the top-level item `transport` under top-level item `cijoe` to match with all other top level identifiers in the configuration-file.

Closes #60

### DIFF
--- a/docs/source/resources/configs/index.rst
+++ b/docs/source/resources/configs/index.rst
@@ -5,7 +5,7 @@ Configs
 
 **cijoe** configuration files are formated using `YAML`_ and named with suffix
 ``.config``. In the core functionality of provided by cijoe, only the key
-``transport`` has special meaning.
+``cijoe.transport`` has special meaning.
 
 Keys are otherwise granted meaning by their use of
 :ref:`sec-resources-scripts`, tests, and regular Python modules.
--- a/src/cijoe/core/command.py
+++ b/src/cijoe/core/command.py
@@ -75,7 +75,7 @@ class Cijoe(object):
         self.transport_local = transport.Local(self.config, self.output_path)
         self.transport = self.transport_local
 
-        ssh = self.config.options.get("transport", {}).get("ssh", None)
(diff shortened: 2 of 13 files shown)


safl (commented):
This probably also need to adjust some ``.config`` files? E.g. the default configs, and possibly some documentation as well?

safl (changes requested):
This probably also need to adjust some ``.toml`` files such as the default configs, and possibly some documentation as well. For example:

* ``src/cijoe/core/configs/transport-ssh.toml``

Let's integrate the merge of the packages https://github.com/refenv/cijoe/pull/63 and then this one after, then there are probably more ``.toml`` files to fix and then it can be done in a single commit.

NaddiNadja:
I did a search for `transport` in the repository, and it looks like only `docs/source/usage/index.rst` and `src/cijoe/core/configs/transport-ssh.toml` needed to be updated, but let me know if you think I'm missing something. :)

safl:
Sounds good, now with https://github.com/refenv/cijoe/pull/63 merged, after you rebase, then there might some additional example-configs to adjust?

safl (approved):

PR #66, opened 2024-06-13
fix(resources): remove unused setuptools import

karlowich:
This import is unused in the code - but it makes installing cijoe on MacOS more painful than it should be.


safl:
This import is necessary due to its side effects, which is why it includes the 'noqa' annotation. Without this import, dynamic loading fails.

safl:
Hmm, seems like I was wrong, so reopening.

Issue #49, opened 2023-01-17
Feature request: Add ansible like variable substitution in workflow.yaml


Panky-codes:
Currently, we can only access the config variables only via worklets using `cijoe.config.options..`.

Having a feature to directly substitute the variable in workflow.yaml itself would be helpful and avoid bolier plate.
Example behaviour:
```
# local.config
fio:
  bin: "/usr/local/bin"
# fio.workflow
- name: Run fio
  run: |
    {{ config.options[fio][bin] }} --io_depth=4 ....
```

safl:
This is a duplicate of: https://github.com/refenv/cijoe/issues/45 closing https://github.com/refenv/cijoe/issues/45 since this current issue has a better description.

safl:
@Panky-codes this was fixed in #45 thus closing it

PR #67, opened 2024-07-15
feat(workflows): add support for constants

NaddiNadja:
To be able to reuse constants throughout the workflow file, let the workflow file have a top-level key called "constants" where a constant can be defined with the &-notation and referenced with an asterisk (*).

### DIFF
--- a/src/cijoe/core/resources.py
+++ b/src/cijoe/core/resources.py
@@ -322,7 +322,7 @@ class Workflow(Resource):
 
         errors = []
 
-        for top in set(topic.keys()) - set(["doc", "config", "steps"]):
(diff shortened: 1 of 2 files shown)


safl:
I think this is great, however, let's let it simmer for a while to see whether the usage you had for this repeats itself, and then, whether the approach you took with naming-on-first-use is generally preferable.

In any case, then this sheds some light on the need for updating the documentation, as doing this is related to https://github.com/refenv/cijoe/issues/49
And the documentation is way out-dated... making a task of here: https://github.com/refenv/cijoe/issues/68

safl:
Closing this for now, let's reopen in case we stumble across it again.

PR #69, opened 2024-08-06
feat(scripts): add powershell support for repository_prep

NaddiNadja:


NaddiNadja:
As we have switched to using bash as default on our Windows machine, this is not needed anymore.

PR #70, opened 2024-08-27
feat(transport): allow specification of shell bin

NaddiNadja:


NaddiNadja:
@safl I was thinking that it might make sense to add some kind of example config for this, but I am unsure whether (a) it's necessary and (b) what it should be named. I don't really like `transport-local.toml`, but on the other hand, the `shell_bin` is used in the `Local` class which extends `Transport`.

karlowich (approved):
LGTM

src/cijoe/core/transport.py
```diff
@@ -34,18 +34,20 @@ def __init__(self, config: Config, output_path: Path):
...
         self.output_ident = "artifacts"
+        self.shell_bin = self.config.options.get("cijoe", {}).get("run", {}).get("shell_bin", "/bin/sh")
 
     def run(self, cmd, cwd, env, logfile):
         """Invoke the given command"""
 
         env = dict(os.environ)
         env.update(env)
 
+        cmd = [self.shell_bin, "-c", cmd]
```
safl, line 45: One concern here, most Unix-shells treat ``-c`` in the same manner. However, seems like e.g. the ``fish`` shell is a bit different, and shells such as "cmd.exe" or powershell, then this would not work at all?

NaddiNadja: I took this from the documentation of `subprocess.Popen`: https://docs.python.org/3/library/subprocess.html#subprocess.Popen

Here it says that if you have `shell=True` (which we had before), it is the equivalent to do `Popen(['/bin/sh', '-c', cmd])`, so I thought it would be able to handle multiple shells?

safl: Yeah, but behind the phrasing "equivalent to" lies that it is not exactly passing ``-c``, rather it is implemented to handle system differences, have a look here:

- https://github.com/python/cpython/blob/ce9f84a47bfbafedd09a25d0f6f0c8209550fb6c/Lib/subprocess.py#L1518
- https://github.com/python/cpython/blob/ce9f84a47bfbafedd09a25d0f6f0c8209550fb6c/Lib/subprocess.py#L1822

Issue #53, opened 2023-03-13
Provide stand-alone execution of cijoe-script


safl:
Current, the way to execute cijoe-scripts are by setting up a workflow. However, it could be made even simpler by:

- [x] Provide a common stand-alone command-line parser for re-use by scripts
- [x] Add examples of this in the docs, starting out with this, as it show the minimal CIJOE functionality
- [x] Add a script-emitter on the ``cijoe`` tool, that is, a command to produce a cijoe-script. Such that all the "boiler-plate" is done and one can simply fill in the actual logic.

A script template could look something like:

```python
#!/usr/bin/env python3
"""
  This is an example of cijoe script

  It has an "special" main() function and a helper-function cli_interface that manages

  The script itself checks whether the hostname of the system under test is "my_machine"
"""
from cijoe.cli import cijoe_cli

def main(args, cijoe, step):
  """This is the main entry-point of a cijoe-script"""

  err, state = cijoe.run("hostname")
  if err:
    return err

  if "my_machine" not in state.output():
    return 1

  return 0

if __name__ == "__main__":
  cli_interface(main)
```

With stand-alone execution of a cijoe-script, then it will be possible to execute a script as the one above by running:

```bash
./my_script.py
```

safl:
Thanks for adding this @NaddiNadja !

PR #71, opened 2024-08-29
feat(cli): provide standalone execution of cijoe-script

NaddiNadja:
Solves #53

@safl I am unsure whether you had another structure in mind. I have tried to reuse as much existing code as possible, but let me know if you have any ideas on how to improve it :)

### DIFF
--- a/src/cijoe/cli/cli.py
+++ b/src/cijoe/cli/cli.py
@@ -4,6 +4,7 @@ import logging as log
 import os
 import shutil
 import sys
+import tempfile
 import time
 from pathlib import Path
 
@@ -81,6 +82,16 @@ def cli_integrity_check(args):
         return errno.EINVAL
 
     return 0
+
+
+def cli_script(args):
+    """Create python script with the boiler-plate code necessary for running cijoe as a standalone script."""
+    filename = "cijoe-script.py"
+    if os.path.exists(filename):
+        sys.exit("Error: File 'cijoe-script.py' already exists.")
+
+    resources = get_resources()
+    shutil.copy(resources["scripts"]["core.example"].path, filename)
 
 
 def cli_resources(args):
@@ -162,6 +173,7 @@ def cli_example(args):
     err = 0
 
     resources = get_resources()
+    print(resources["configs"].keys())
 
     resource = resources["configs"].get(f"{args.example}.default-config", None)
     if resource is None:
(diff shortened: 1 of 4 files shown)


src/cijoe/cli/cli.py
```diff
@@ -83,6 +84,28 @@ def cli_integrity_check(args):
     return 0
 
 
+def cli_script(args):
+    """Create python script with the boiler-plate code necessary for running cijoe as a standalone script."""
+    filename = "cijoe-script"
+    i = 0
```
safl, line 90: I think it would be preferable if this would always use the same filename, and then print an error and exit in case the script exists.

safl: E.g. something like:

```
with Path("cijoe-script.py") as path:
  if path.exists():
    output error-message
    return ....
    
   path.write_text(content)
```

docs/source/usage/cijoe-script.py
```diff
@@ -0,0 +1,7 @@
+from cijoe.cli.cli import cli_interface
+
+def main(args, cijoe, step):
+    cijoe.run("echo 'Hello World!'")
```
safl, line 4: Can you expand this with:

```python
err, _ = cijoe.run("echo 'Hello World!'")
return err
```

I don't have a great example right now, but it would be really nice for the this "skeleton" script to show what one might commonly do, so perhaps even something like:

```python
err, state = cijoe.run("echo 'Hello World!'")

# Do this to demonstrate how to process command-output
if "Hello" not in state.output():
   ... show some logging here ...
   
return err
```

What do you think?

src/cijoe/cli/cli.py
```diff
@@ -83,6 +84,29 @@ def cli_integrity_check(args):
...
+        """    if "Hello" not in state.output():""",
+        """        return 1""",
+        """    return err""",
+        "",
+        """if __name__ == "__main__":""",
+        """    cli_interface(__file__)""",
+    ]
+
+    with open(filename, "x") as script:
+        script.write("\n".join(lines))
```
karlowich, line 107: How about doing something similar to `cli_example()`, i.e. having an example script in `src/cijoe/core/scripts` and copying that. That way you don't have to write the code as strings.
Also that would allow the documentation to just point to that script when describing the output.

karlowich (commented):
One suggestion

src/cijoe/cli/cli.py
safl: Good idea @karlowich !

NaddiNadja: @safl can you check that this is the "right" way to include it in this way?

safl: > @safl can you check that this is the "right" way to include it in this way in the documentation?

The documentation needs an overhaul, so just getting the information in there is great, so, LGTM!

src/cijoe/cli/cli.py
```diff
@@ -83,6 +84,18 @@ def cli_integrity_check(args):
     return 0
 
 
+def cli_script(args):
+    """Create python script with the boiler-plate code necessary for running cijoe as a standalone script."""
+    filename = "cijoe-script.py"
+    if os.path.exists(filename):
+        sys.exit("Error: File 'cijoe-script.py' already exists.")
+
+    cijoe_path = str(cijoe.__path__[0])
```
safl, line 93: At this point you could now go all the way and do a simple file-copy, e.g. ``shutil.copyfile()``.
To locate the script itself, then you can place the script in ``cijoe/src/cijoe/core/scripts/``, and then you can rely on the regular resource-locator.

safl (approved):
Excellent!

PR #72, opened 2024-09-03
fix(core): add script parent path to sys.path

NaddiNadja:
Since cijoe-scripts are dynamically loaded, the current work directory is not available as import path. Therefore, we inject the parent path to sys.path in order to allow imports of scripts from the same directory.


safl (approved):
Great!

PR #74, opened 2024-09-19
Stand-alone execution

NaddiNadja:
Allows cijoe-scripts to be run with cijoe, so for example `cijoe cijoe-script.py`. The `.py` suffix is reserved for python scripts, and cannot be used in workflow step names. 

Multiple cijoe-scripts can be executed in sequence as a workflow, for example `cijoe script1.py script2.py`, but requires that all arguments ends with `.py`. If one or more arguments does not end with `.py`, the arguments are treated as workflow steps.

### DIFF
--- a/src/cijoe/cli/cli.py
+++ b/src/cijoe/cli/cli.py
@@ -1,5 +1,6 @@
 import argparse
 import errno
+import jinja2
 import logging as log
 import os
 import shutil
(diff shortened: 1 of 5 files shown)


src/cijoe/cli/cli.py
```diff
@@ -351,27 +351,37 @@ def cli_workflow(args):
...
 
 
-def cli_interface(path):
-    path = Path(path)
-    args = parse_args()
-
+def create_tmp_workflow(args, paths):
     with tempfile.NamedTemporaryFile() as workflow:
         setattr(args, "workflow", Path(workflow.name))
+        setattr(args, "step", [])
```
safl, line 357: How about putting this into a jinja2 template? I think there is some general infra-structure in the code this.

safl (commented):
Can you wrap the commit-messages to 79 chars line-width?

src/cijoe/cli/cli.py
```diff
@@ -351,27 +352,33 @@ def cli_workflow(args):
...
 
 
-def cli_interface(path):
-    path = Path(path)
-    args = parse_args()
+def create_tmp_workflow(args, paths):
+    resources = get_resources()
+    paths = map(Path, paths)
+
+    template_path = resources["templates"]["core.example-tmp-workflow.yaml"].path
```
safl, line 359: Let's call it ``adhoc`` instead of **tmp**, since **tmp** is such as  loaded term...

safl (commented):
The commit-message description lacks a description of the issue that this solves. E.g. that the Python script needs to be executed with the Python interpreter that has the virtual environment loaded in which cijoe is installed. Just a "traditonal":

```
python3 cijoe-script.py
```

Will most likely fail, since the above implies that cijoe is installed system-wide which is rarely possible since PEP-668: https://peps.python.org/pep-0668/

Btw. this is awesome, thanks! :)

NaddiNadja:
@safl, i basically stole your wording for the commit message, but it's been changed now :)

PR #77, opened 2024-09-24
feat(core): add put and get scripts

NaddiNadja:
Added core scripts for the put and get interface too provide predefined workflow steps for copying files between local and remote


safl (approved):
Excellent!

Issue #73, opened 2024-09-18
Stand-alone execution


safl:
The recently added functionality with standalone execution of CIJOE scripts assume that the user knows which she-bang to use or calls the script with e.g.:

```python
$HOME/.local/pipx/venvs/cijoe/bin/python3 cijoe-script.py
```

Or with the absolute path from above resolved and added as a she-bang inside of ``cijoe-script.py``.

This is needed such that the **cijoe** modules are actually loadable from the script. Thus, the following should be added.

# Nice to have

- [x]  ~~Prepend ``sys.executable`` as a she-bang in the script emitted by``cijoe --script``~~
- [x] Make it possible to execute cijoe-scripts **with** cijoe, e.g. ``cijoe cijoe-script.py``
- [x] Drop ``--script`` cli-option and instead emit the script along with the examples when calling ``cijoe --example``, then each **cijoe package** can provide example-scripts which is very neat
- [x] Check, ahead-of-time, that the scripts given to ``cijoe one/path/script1.py another/path/script2.py`` are within **reach** of the cijoe auto-discovery. E.g. if the user provides ``cijoe /tmp/foo/jazz/script1.py`` then it will currently fail. Alternatively, then the scripts should be moved to some other location, within scope, when given as cli-arguments.

# Needs some additional thought

- [x] ~~Add emitter of script-shebang ``cijoe --script-shebang``, should output something like ``#!/path/to/the/python/currently/in/use`` which the user can either manually add to the script, if they want, or use the path to execute the script, e.g. ``/path/to/the/python/currently/in/use cijoe-script.py``~~

safl:
Now that the search path is expanded when doing ``cijoe path/to/jazz/script.py``, how about dropping the need to explicitly call the``cli_interface`` helper-function inside of the script? This will reduce the stand-alone execution use-case to:

```python
cijoe path/to/jazz/script.py
```
Dropping support for:

```python
# Use python interpreter directly
python path/to/jazz/script.py

# Rely on "#!" / shebang
./path/to/jazz/script.py
```

The intent here is to simplify usage, then regardless of whether one is executing scripts of workflows, then the entry-point is the same; the **cijoe** comand-line. Also, it arguably makes the scripts "simpler".

NaddiNadja:
[...]

PR #79, opened 2024-09-24
fix(cli/script): emit example script when calling `cijoe --example`

NaddiNadja:
Instead of having the `cijoe --script` cli-option, emit the example script along with example config and workflow when calling `cijoe --example`.

Solves point 3 under "Nice to have" in #73


safl (approved):
Excellen!

PR #80, opened 2024-09-24
fix(cli): add standalone script paths to search paths for auto-discovery 

NaddiNadja:
When running cijoe-scripts with the `cijoe` cli, they might be out of reach for the cijoe auto-discovery. In order to allow execution of the scripts, add parent paths to collected paths.

Solves point 4 under "Nice to have" in https://github.com/refenv/cijoe/issues/73


safl (approved):
Excellent!

Issue #84, opened 2024-10-01
Experiment with async / immediate output-processing via SSH


safl:
Currently commands are invoked via Paramiko exec_command() and it waits until the command is done executing before flushing it to run.log. It would be excellent with the following enhanced behavior:

- [ ] Stream the output from command execution
- [ ] Dump output to run.log and optionally also to stdout

This should be able to replace the current "monitor" functionality, and thus, could take over the use of the ``--monitor`` argument.

By doing so, then **cijoe** can provide a "native" feel when executing commands, even when these commands are remote. There are starvation issues to consider when implementing this, but it would be excellent.

PR #85, opened 2024-10-10
monitoring: add real-time monitoring of workflow output 

NaddiNadja:
We do not use the functionality "monitor" anymore. To allow us to use the `--monitor` argument for monitoring the workflow output in real time, we remove the old funcitonality. 
Then, stream the output from the subprocess (in `Local.run()`) and from paramiko (in `SSH.run()`) to allow for real-time monitoring and report generation with partial output.

Solves issue #84


src/cijoe/cli/cli.py
```diff
@@ -428,6 +428,12 @@ def parse_args():
         const=1,
         help="Increase log-level. Provide '-l' for info and '-ll' for debug.",
     )
+    workflow_group.add_argument(
+        "--monitor",
+        "-m",
+        action="store_true",
+        help="Monitor workflow-output in the current-workdir-directory (cwd).",
```
safl, line 435: Perhaps rephrase this to something along the lines of: "Dump command output to stdout"?

src/cijoe/core/command.py
```diff
@@ -104,20 +108,27 @@ def _run(self, cmd, cwd, env, transport):
...
-                err=err,
-                begin=begin,
-                end=time.time(),
+                err=0,
+                begin=0,
+                end=0,
                 output_dpath=cmd_output_dpath,
                 output_fpath=cmd_output_fpath,
                 state_fpath=cmd_state_fpath,
+                done=False,
```
safl, line 120: Perhaps ``is_running`` or ``is_done`` is preferable?

src/cijoe/core/command.py
```diff
@@ -104,20 +108,27 @@ def _run(self, cmd, cwd, env, transport):
...
             )
             state.to_file()
 
+            begin = time.time()
+            err = transport.run(cmd, cwd, env, logfile, self.monitor)
```
safl, line 125: Why is it required to move the creation of the **CommandState** before the execution?

src/cijoe/core/transport.py
```diff
@@ -35,20 +35,26 @@ def __init__(self, config: Config, output_path: Path):
...
         """Invoke the given command"""
 
         env = dict(os.environ)
         env.update(env)
 
         with subprocess.Popen(
             cmd,
-            stdout=logfile,
-            stderr=subprocess.STDOUT,
+            stdout=subprocess.PIPE,
```
safl, line 46: Don't you need to also redirect stderr to stdout for this to combine the two streams?

safl (commented):
This is so very exciting! I have a couple of questions inline, also I am a bit concerned about starvation due to past experience (https://stackoverflow.com/questions/3575554/python-subprocess-with-timeout-and-large-output-64k) however it mi might be completely irrelevant today and we can just give it a spin and experiment with it, perhaps a test could be added that generated "alot" of data e.g. emitting megabytes of text, testing it with and without newlines in the "selftest".

You can also add the removal of **watchdog** from project dependencies:
https://github.com/refenv/cijoe/blob/main/setup.cfg

src/cijoe/core/transport.py
NaddiNadja: That depends: do we want `stderr` to be printed as output in the workflow report? With the setup here, it's only printed in the initiator's terminal, but if I set `stderr=subprocess.STDOUT`, both `stderr` and `stdout` will be printed as output in the workflow report.

safl:
Yeah, it would be great for stderr to be included.

tor. 10. okt. 2024 16.22 skrev NaddiNadja ***@***.***>:

src/cijoe/core/command.py
NaddiNadja: I need to call the `state.to_file()`, such that we step-data for the workflow report, if we generate it mid step (with `cijoe -p`)

src/cijoe/core/transport.py
safl: Yup, stderr should be part of output. That is how it is now collected in the various ``run.log`` files.
I was thinking it might be nice to do something like below, inspired by the ``tee`` command.
That is, implement something "file-like" and "with-usable" such that it can be passed to Popen.

```python
class Tee:
    def __init__(self, file_path):
        self.file = open(file_path, 'w')

    def write(self, data):
        self.file.write(data)
        self.file.flush()

        sys.stdout.write(data)
        sys.stdout.flush()

    def flush(self):
        self.file.flush()

    def close(self):
        self.file.close()

    def __enter__(self):
        return self

    def __exit__(self, exc_type, exc_value, traceback):
        self.close()

with Tee('run.log') as tee:
  # Set up the subprocess with stderr redirected to stdout
  process = subprocess.Popen(
    cmd,
    stdout=tee,
    stderr=subprocess.STDOUT,
    ...
  )
```

safl: Hmm, might even be added to the state class?

safl: Hmm, nevermind, this was a bad suggestion since ``subprocess.Popen()`` does not support "file-like".
The approach of:

```
    stdout=subprocess.PIPE,
    stderr=subprocess.STDOUT,
```

Seems preferable.

NaddiNadja: As we talked about, one problem is that using `subprocess.Popen(..., stdout=tee,...)`, `Popen` will not use the file-like's `write()` method for writing, according to [this StackOverflow thread](https://stackoverflow.com/questions/21035127/how-to-use-a-custom-file-like-object-as-subprocess-stdout-stderr).

I have solved it by modifying the `write()` method of `Tee` and using `PIPE` in `Popen`

PR #87, opened 2024-10-15
bugfix(transport): combine stdout and stderr instead of using get_pty

NaddiNadja:
`get_pty=True` must have some other side-effects than anticipated, as if `cmd` is set to for example `powershell "echo hi"`, then `stdout.channel.exit_status_ready()` will never return True. The parameter is not necessary for streaming output in real-time though, so we combine `stdout` and `stderr` with the `set_combine_stderr(True)` instead.


safl (approved):
Nice catch!

PR #89, opened 2024-10-15
Cleanup more

safl:


.pre-commit-config.yaml
```diff
@@ -40,4 +40,4 @@ repos:
     # F401 = F811; This happens for fixtures for indirect parametrize
     args:
     - --max-line-length=88
```
karlowich, line 42: Are you sure you want to ignore line-lengths?
If yes, I guess we don't need a max length

src/cijoe/core/resources.py
```diff
@@ -45,21 +45,25 @@
...
 from pathlib import Path
+from typing import Any, Dict, List, Union
 
 try:
     from importlib.resources import files as importlib_files
 except ImportError:
-    from importlib_resources import files as importlib_files
+    from importlib_resources import files as importlib_files_fallback
+
+    importlib_files = importlib_files_fallback
```
karlowich, line 55: Is this approach really better? Both are confusing and hard to read, but at least the other is simpler

karlowich (commented):
Two questions

src/cijoe/core/resources.py
safl: It is what mypy prefers.

pyproject.toml
```diff
@@ -1,15 +1,74 @@
...
+
+[tool.setuptools.packages.find]
+where = ["src"]
+namespaces = true
+
+[tool.setuptools.dynamic]
+version = { attr = "cijoe.core.__version__" }
 
-[tool.isort]
-profile = "black"
```
mbrsamsung, line 12: Why are we removing the isort configuration?

.pre-commit-config.yaml
safl: flake8 uses ``--max-line-length=88`` and E501 for different things. I don't want the tooling to choke on flake8 complaining about a url in documentation string being too long. Instead, we can rely on **black** for fixing line-length / format in a saner maner, and then rely on flake8 to take care of the language integrity things that **black** does not cater to.

mbrsamsung (commented):
Got a question

pyproject.toml
mbrsamsung: nm i can see you answered the same ish question that Karl asked

mbrsamsung (approved):
lgtm

pyproject.toml
safl: I guess it might be possible to replace flake8 with ruff and then put all tool configs into pyproject.toml, that would be nice.

mbrsamsung: Its a possibility, we can look into it if you like

safl: That would be nice!

safl: @mbrsamsung replaced flake8 with ruff and moved tool configs into ``pyproject.toml``. There is now a single redundant config option for mypy. But reaching 0 is probably not possible. Anyway, this should now enable misc. editors to make use of the tool config, which is a nice addition :)

src/cijoe/qemu/wrapper.py
```diff
@@ -114,10 +114,10 @@ def is_up(self, timeout=120):
...
             enter = time.time()
             try:
-                with self.serial.open() as serialfile:
-                    if "login:" in serialfile.read():
-                        time.sleep(10)
-                        return True
+                if "login:" in self.serial.read_text(
+                    encoding="utf-8", errors="replace"
+                ):
+                    return True
```
karlowich, line 120: This is awesome, we've had that issue for a loooong time

karlowich (approved):

PR #90, opened 2024-10-17
fix(cli): drop support for running scripts with python interpreter

NaddiNadja:
Now that standalone cijoe scripts can be run with
`cijoe path/to/cijoe-script.py`, we drop support for running standalone cijoe scripts with the python intepreter. This way we only have one entrypoint of running cijoe-scripts, which ultimately makes both the cijoe scripts and cijoe as a tool simpler.

Solves the comment from Oct 16 2024 on issue #73


safl (approved):
Excellent!

Issue #75, opened 2024-09-23
Multiple transports in a single configuration


safl:
It is not uncommon that one needs to utilize multiple transport endpoints, however, since **cijoe** does not support this, then there are awkward limitations. E.g. when setting up qemu-guests, then these commands should be able to run on any transport, but only "local" is possible, this is because the "ssh" transport is utilized to SSH into the machine itself.
Thus, it would be great to support multiple user-defined transports.

NaddiNadja:
This has been fixed in #94

PR #94, opened 2024-10-24
feat(transport): allow defining multiple transport endpoints

NaddiNadja:
Added the `endpoint_name` parameter to `cijoe.run(...)`, which allows users to specify which endpoint they want to execute commands on. Endpoints are defined in the configuration file under the key `cijoe.transport` with user-defined names. If the `endpoint_name` parameter is not given, the command is executed on the first defined endpoint. If no endpoints are defined, the command is executed on the initiator.

Solves issue #75

### DIFF
--- a/src/cijoe/core/command.py
+++ b/src/cijoe/core/command.py
@@ -151,13 +151,20 @@ class Cijoe(object):
         os.makedirs(os.path.join(self.output_path, self.output_ident), exist_ok=True)
 
         self.transport_local = transport.Local(self.config, self.output_path)
-        self.transport = self.transport_local
+        self.transports = {}
 
-        ssh = self.config.options.get("cijoe", {}).get("transport", {}).get("ssh", None)
-        if ssh:
-            self.transport = transport.SSH(self.config, self.output_path)
+        config_transports = self.config.options.get("cijoe", {}).get("transport", {})
+        for transport_name in config_transports:
+            self.transports[transport_name] = transport.SSH(
+                self.config, self.output_path, transport_name
+            )
 
-    def set_output_ident(self, output_ident: str):
+        if not config_transports:
+            # If there are no endpoints defined, create one transport which is the
+            # local transport.
+            self.transports["local"] = self.transport_local
+
+    def set_output_ident(self, output_ident: str, transport_name=None):
         """
         This sets the output-identifier which is used in order to provide a subfolder
         for artifacts, command-output etc. Additionally, then it reset the command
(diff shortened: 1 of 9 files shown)


karlowich (commented):
I propose that we rename `endpoint_name` to `endpoint`. To me this is cleaner, and makes for simpler code e.g. `for endpoint in endpoints:`

NaddiNadja:
> I propose that we rename `endpoint_name` to `endpoint`. To me this is cleaner, and makes for simpler code e.g. `for endpoint in endpoints:`

you're probably right, i have fixed this now, and moved the logic for finding the transport to a single method to reduce code redundancy

src/cijoe/core/command.py
```diff
@@ -196,7 +203,17 @@ def _run(self, cmd, cwd, env, transport):
 
         return err, state
 
-    def run(self, cmd, cwd=None, env={}):
+    def _get_transport(self, endpoint=None):
+        if not endpoint:
+            endpoint = next(iter(self.transports))
+        if endpoint not in self.transports:
```
karlowich, line 209: Maybe this should be an `else if` since if the first is condition is true, this will always be true as well.
I initially misread the code, perhaps this change could make it a bit clearer.

safl:
> > I propose that we rename `endpoint_name` to `endpoint`. To me this is cleaner, and makes for simpler code e.g. `for endpoint in endpoints:`
> 
> you're probably right, i have fixed this now, and moved the logic for finding the transport to a single method to reduce code redundancy

In the issue #75, I referred to it as an "endpoint" / "transport endpoint," but this was likely not ideal, as it is referred to as "transport" everywhere else. I propose renaming it to ``transport_name: str`` to indicate that this is not an instance of the ``Transport`` class but rather a name specified in the configuration file.

safl (changes requested):
The commit message is now out of "tune" with the implementation and docs.

src/cijoe/core/command.py
NaddiNadja: I know that if line 207 evaluates to true, then 209 will always evaluate to false (as the code is now), but because we change the value of `endpoint` (now renamed to `transport_name`) in line 208, I prefer to use another `if` over an `elif`. If we mess with line 208 in the future, line 209 is there to ensure that it doesn't break.
It's a constant-time lookup, so performance wise, it makes no difference, so I believe it's just a matter of preference.

safl (approved):
LGTM.

karlowich (approved):

PR #96, opened 2024-10-28
feat: add support for failing fast

safl:


src/cijoe/core/command.py
```diff
@@ -268,7 +268,7 @@ def getconf(self, key: str, default: Any = None):
         envvar = os.getenv(envkey)
         if envvar:
             log.debug(f"found {key} ({envkey}) in environment variables.")
-            return envvar
+            return convert_str(envvar)
```
karlowich, line 271: This should probably be added to the doc-string. To me it is not entirely logical that this is the behaviour

docs/source/workflows/index.rst
```diff
@@ -23,6 +23,27 @@ similar to how targets are specified in a :make:`Makefile <>`:
...
+which you can set in your :ref:`sec-resources-configs`:
+
+.. literalinclude:: ../../../src/cijoe/core/configs/default-config.toml
+
+By default **cijoe** functionality, then this can also be controlled environment
+variable:
+
+.. code-block:: bash
+
+   CIJOE_FAIL_FAST=1
```
karlowich, line 45: I think it's better to have `CIJOE_FAIL_FAST=true` as the example. I'm sure both work - but given that the code in `cli.py` uses `False` as the default value, I think we should encourage users to also use booleans

karlowich (changes requested):

src/cijoe/core/command.py
safl: When you view ``getconf()`` as a method for retrieving values from configuration files with environment variable overrides, converting ``.toml`` values from strings to booleans or integers makes sense. This change ensures scalar values are treated the same, aligning with the configuration retrieval behavior, rather than returning unconverted values.

Issue #83, opened 2024-10-01
Add checksum checks on qemu images


safl:

NaddiNadja:
This was fixed with #100

PR #100, opened 2024-10-31
feat(qemu): add checksum checks on qemu images

NaddiNadja:
Before, images were not downloaded if a file on the given path already existed. Now, checksums are compared to determine whether the image should be downloaded. Configuration files must be expanded with a qemu.guest.<init>.url_csum which points to the checksum file of the downloaded image.

Solves issue #83

----

@safl, I create this as a draft for now. I still need to update the example configuration files with the `url_csum`, but as I don't have access to Digital Ocean, I think I need your help on this one. 

Also, I don't know if I'm 100% satisfied with the name of the `_get_img()` method - maybe you have an idea? The method contains all logic for checking checksums and downloading the new image if necessary, and then returns the path to the newly-downloaded or already-existing image.

### DIFF
--- a/src/cijoe/qemu/scripts/guest_init_using_bootimage.py
+++ b/src/cijoe/qemu/scripts/guest_init_using_bootimage.py
@@ -11,6 +11,8 @@ Config::
 
     [qemu.guest.init_using_bootimage]
     url = # URL pointing to download location of the bootable disk-image
+    url_checksum = # URL pointing to download location of the checksum of the 
+                   # bootable disk-image
     img = # Absolute path to disk-image file"
 
 Retargetable: False
--- a/.github/workflows/selftest.yml
+++ b/.github/workflows/selftest.yml
@@ -16,15 +16,33 @@ defaults:
 
 jobs:
   format:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        python-version: ['3.12']
 
     steps:
     - name: Grab source
       uses: actions/checkout@v4.1.7
 
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v5.2.0
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Python dependencies and their versions
+      run: |
+        brew update || true
+        brew upgrade pipx || true
+        make deps
+        make info
+        echo "$HOME/.local/bin" >> $GITHUB_PATH
+
     - name: Install pre-commit
       run: |
-        pipx install pre-commit
+        pipx install pre-commit --python ${{ matrix.python-version }}
 
     - name: Run pre-commit
       run: |
--- a/src/cijoe/core/misc.py
+++ b/src/cijoe/core/misc.py
@@ -1,5 +1,9 @@
 import errno
+import hashlib
+import logging as log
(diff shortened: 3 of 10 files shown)


src/cijoe/qemu/scripts/guest_init_using_bootimage.py
```diff
@@ -11,6 +11,8 @@
 
     [qemu.guest.init_using_bootimage]
     url = # URL pointing to download location of the bootable disk-image
+    url_csum = # URL pointing to download location of the checksum of the
```
safl, line 14: Let's call it ``url_checksum``

safl (commented):
I think this is a good start. Would it make sense to add the logic to the ``download()`` and expand it with the checksum logic?

NaddiNadja:
> I think this is a good start. Would it make sense to add the logic to the `download()` and expand it with the checksum logic?

I think it would then be better to have a `download_with_checksum()` or something like that? Because it needs to have an extra parameter for the checksum URL, and it would need to call `download()` twice as it does now. Would that make sense?

safl:
> > I think this is a good start. Would it make sense to add the logic to the `download()` and expand it with the checksum logic?
> 
> I think it would then be better to have a `download_with_checksum()` or something like that? Because it needs to have an extra parameter for the checksum URL, and it would need to call `download()` twice as it does now. Would that make sense?

Sure, that makes sense, possibly ``download_and_verify()``, alternatively since it is Python, then adding an additional optional argument would not break the current usage of the function, you could have e.g. ``download(..., url_checksum=None)``. And then document the behavior in the doc-string, that when giving a check-sum url. But yeah... just adding another function might be "cleaner"... sure go with that :)

NaddiNadja:
Hi @safl, I have ended a place now where the code is a much simpler than before. 

I had to take into account that different sources might use different algorithms for checksums. I have made the assumption that the algorithm is always part of the filename (case-insensitive) - is that a valid assumption? 

I use `hashlib` to loop over valid algorithms and also use that for finding the checksum of the files with `hashlib.file_digest()`. The formatter apparently think that `hashlib` doesn't have this method, so I don't know how to get it to pass that test? Locally, `make format-all` doesn't complain (and it also works fine when I test it), so maybe there is something with versioning?

I have split the "download checksums + find algorithm in filename" to a separate function, because it felt cleaner to me. What do you think about that?

safl (commented):
Would it be sensible to simplify and rely on the CLI tool ``sha256sum``? As far as I know, ``sha256`` is used by most hosts of ``.iso``, ``.img``, and ``.qcow2`` files, such as Fedora, Debian, Ubuntu, and FreeBSD.

By using a CLI tool, this part of the code is prepared in case we decide to make the qemu script "re-targetable." At that point, we could also incorporate either ``curl`` or ``wget`` for downloads, enabling the script to run from locations other than the initiator. This capability could be very useful for qemu.

NaddiNadja:
@safl, I just looked at the example configs for images, and for Debian they use sha512sums

https://cloud.debian.org/images/cloud/bookworm/daily/latest/
https://cloud.debian.org/images/cloud/bullseye/daily/latest/

safl:
> @safl, I just looked at the example configs for images, and for Debian they use sha512sums
> 
> https://cloud.debian.org/images/cloud/bookworm/daily/latest/
> https://cloud.debian.org/images/cloud/bullseye/daily/latest/

You are right. I guess the upside it is still "sha".

PR #108, opened 2024-12-09
Add the creation of a Docker image that can be used by cijoe itself.

safl:
To test the cijoe QEMU package and the system-imaging packages, they should be verified using GitHub workflows. This preparation involves providing a Docker image with QEMU installed, which can be used in a GitHub workflow.

There is also a fix in here as I stumbled upon it.

### DIFF
--- a/.github/cijoe-docker/Dockerfile
+++ b/.github/cijoe-docker/Dockerfile
@@ -1,0 +1,59 @@
+FROM debian:bookworm
+
+WORKDIR /opt
+
+RUN apt-get -qy update && \
+	apt-get -qy \
+		-o "Dpkg::Options::=--force-confdef" \
+		-o "Dpkg::Options::=--force-confold" upgrade && \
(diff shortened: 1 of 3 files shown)


.github/qemu-docker/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+	git \
+	htop \
+	lshw \
+	neovim \
+	openssh-server \
+	pipx \
+	procps \
+	python3-build \
+	python3-jinja2 \
+  qemu-kvm \
```
NaddiNadja, line 25: nitty gritty, but should the indentation be the same here?

NaddiNadja (approved):
lgtm, there is just one teenie tiny thing

.github/qemu-docker/Dockerfile
safl: Nice catch!

PR #111, opened 2024-12-10
Add coverage of tests and refactor workflows for coverage report

mbrsamsung:


.github/workflows/linux.yml
```diff
@@ -0,0 +1,64 @@
+---
+name: linux
```
safl, line 2: Instead of adding this file, then you can expand the matrix in:

``.github/workflows/cijoe_packages.yml``

example:

```
    strategy:
      fail-fast: false
      matrix:
        usage_example: ['fio', 'linux', 'system_imaging', 'qemu.build']
        python-version: ['3.12']
```

safl (changes requested):
See inline.

.github/workflows/cijoe_packages.yml
```diff
@@ -52,7 +52,15 @@ jobs:
 
     - name: Run it!
       run: |
-        cijoe --monitor -l
+        coverage run\
+          --data-file=cijoe-output/.coverage\
+          --rcfile=repository/.coveragerc\
+          -L \
+          $(which cijoe) --monitor
+        coverage\
```
safl, line 60: How about moving the "coverage report" down to the collection-stage? Then there is a stage which run cijoe and produces the cijoe-report.

.github/workflows/cijoe_packages.yml
```diff
@@ -52,7 +52,15 @@ jobs:
 
     - name: Run it!
       run: |
-        cijoe --monitor -l
+        coverage run\
+          --data-file=cijoe-output/.coverage\
+          --rcfile=repository/.coveragerc\
```
safl, line 57: How about placing the coverage files in their own artifact? Currently, downloading the cijoe-report also includes the coverage info, which might give the impression that including coverage info is a standard cijoe behavior

safl (changes requested):
This is growing in different directions, perhaps wrap up it up in smaller PRs for incremental improvements?

safl:
Closing this as the changes were rebased and extended on https://github.com/refenv/cijoe/pull/143

PR #133, opened 2024-12-13
change(gitignore): Add vscode and venv to git ignore

mbrsamsung:
Vscode setttings directory is deleted every time `make all` is executed. This would fix it. The same goes for virtual environment


safl (approved):
lgtm

PR #136, opened 2024-12-13
fix(docker): remove errornoeous `--include-deps` flag from pipx install

mbrsamsung:
The `--include-deps` flag was mistakenly added to the pip install(pipx) command for the pytest-cov package


safl (commented):
I guess this: "The `--include-deps` flag was mistakenly added to the pip install
command for the pytest-cov package"
is the "pipx inject" command?

mbrsamsung:
Correct, it is the pipx injection command

PR #137, opened 2024-12-13
Adding docker build makefile job and fix "fixup docker"

mbrsamsung:
To be able to locally test building of docker container a new `docker-build` command has been added to the makefile.

This enables local debug of the docker container build process. This is needed to fix the issue that was caused by missing ampersand for splitting multi-line run commands in docker.

The second commit adds these missing ampersand and make the docker container build again.


Makefile
```diff
@@ -8,6 +8,7 @@ PYTEST="$(shell pipx environment -V PIPX_LOCAL_VENVS)/${PROJECT_NAME}/bin/pytest
 PYTHON_SYS=python3
 PYTHON_VENV="$(shell pipx environment -V PIPX_LOCAL_VENVS)/${PROJECT_NAME}/bin/python3"
 TWINE=twine
+CIJOE_VERSION=v0.9.45
```
safl, line 11: How about this instead:
```
CIJOE_VERSION=v$(shell cd src; python3 -c "from cijoe import core; print(core.__version__)")
```

safl: I force-pushed this suggestion.

Issue #128, opened 2024-12-12
Improve general test and verification of cijoe


safl:
The current test suite for `cijoe` includes a basic `pytest` implementation that exercises various aspects of the project, such as:

- Workflow loading
- CLI behavior
- Core package scripts (indirectly)

These tests primarily focus on unit-testing the core scripts, which implement the command-runner and other critical functionalities. However, a significant portion of the codebase remains untested.

# Tasks

1. **Expand Test Coverage:** setup GHA workflow exercising all the cijoe package examples
- [x] Emit the example e.g. `cijoe --example qemu`
- [x] Run it `cijoe --monitor -ll`
  - These examples are what one turns to, to see how to use a package, thus, it is critical that they are actually tested
  - Currently done here: `.github/workflows/cijoe_packages.yml`, now refactored/merged with "selfttest" the new workflow is now ``verify_and_publish``
- [x] Add run of all examples (see coverage reports for those missing)
3. **Introduce Coverage Reports:** Integrate a coverage tool to track and enforce coverage metrics.
- [x] Combine all the coverage files to produce a coverage report available as a downloadable artifacts
- [x] Publish coverage files to https://coveralls.io/
  - [x] https://coveralls.io/github/refenv/cijoe 
  - [x] Fix source-links, in the report, then currently one has to navigate through the pipx-venv path which breaks linkage to project source. Thus, this must be fixes, such that it is easy to jump to the code which have not yet been tested properly
- [x] Add badge in cijoe repository with link to coverage report

5. **Refactor Tests if Necessary:**
  - Many of the examples require split of workflows and configurations into smaller parts.
6. Document the packages
  - When documenting the Python package and modules then the text is updated here: https://cijoe.readthedocs.io/en/latest/packages/index.html
  - Thus, "simply" document the modules in a form that is sensible for users of cijoe

# Suggested Tools

- Coverage.py https://coverage.readthedocs.io/en/7.6.9/
  - For measuring code coverage and offline reporting
  - Example usage: ``coverage run ...``
- Coveralls
  - For online reporting
  - https://coveralls.io/

# Related issues and PRs

- https://github.com/refenv/cijoe/issues/115
- https://github.com/refenv/cijoe/issues/124
- https://github.com/refenv/cijoe/issues/126

- https://github.com/refenv/cijoe/pull/145
- https://github.com/refenv/cijoe/pull/134
- https://github.com/refenv/cijoe/pull/129
- https://github.com/refenv/cijoe/pull/127
[...]

PR #144, opened 2025-01-06
Part of #128: Fix issues with paths of combined coverage report

mbrsamsung:
This PR is part of issue #128 

This will produce a combined coverage report that will be used in PR branches and when merging into the main branch for publishing.


safl:
As discussed offline, then having tests in-module was a blocker.
Thus, have a look here: https://github.com/refenv/cijoe/pull/145
Let's get that in first, then rebase #144 on top to see whether it improved.

tests/core/__init__.py
safl, line 1: I don't see a reason to add these ``__init__.py`` files, pytest do not need them, it will just confuse it, thinking they are python packages, which they are not.

pyproject.toml
```diff
@@ -88,3 +88,24 @@ lint.extend-ignore = ["E203", "F401", "F811", "E501"]
 [tool.mypy]
 ignore_missing_imports = true
 exclude = "docs/source/conf.py"
+
+[tool.coverage.run]
```
safl, line 92: Looks like this option is missing: ``branch = True``, it was there previously to include tracking of whether branches are taken.

mbrsamsung: Don't ask why it is needed to have the --cov-branch flag added to the pytest invocation. I have no idea why it fails when its only added as a pyproject.toml config parameter.

safl (approved):
Awesome!

PR #146, opened 2025-01-09
Add null_blk example in the CI pipeline

mbrsamsung:


Makefile
```diff
@@ -54,6 +54,7 @@ docker:
 	@docker run -it \
 				-w /tmp/source \
 				--mount type=bind,source="$(shell pwd)",target=/tmp/source \
+				--device=/dev/kvm \
```
safl, line 57: This is problematic for those on Mac. Does it do something that ``make docker-privileged`` cannot cover?

safl (commented):
Minor thing inline.

safl (commented):
Looks good, a couple of questions inline.

Makefile
mbrsamsung: was removed again

PR #147, opened 2025-01-09
feat(tests): test for various cli parameter invocations

mbrsamsung:


src/cijoe/cli/cli.py
```diff
@@ -182,8 +182,8 @@ def cli_produce_report(args):
 
 def cli_example(args):
     """Create example config.toml and workflow.yaml"""
-
     log.info("cli: examples")
+    print(f"args -> {args.example}")
```
safl, line 186: Is this print intentional?

tests/core/test_cli_example.py
```diff
@@ -1,3 +1,9 @@
+import os
+from pathlib import Path
```
safl, line 2: Is ``Path`` ever used? I wonder why format is not complaining about it.

safl (commented):
A couple of questions inline.

.github/workflows/verify_and_publish.yml
```diff
@@ -111,7 +111,7 @@ jobs:
         - qemu.guest_x86_64
         - system_imaging.aarch64
         - system_imaging.x86_64
-        python-version: ['3.12']
+        python-version: ['3.11']
```
safl, line 114: Why this?

.github/workflows/verify_and_publish.yml
```diff
@@ -89,7 +89,7 @@ jobs:
 
   examples:
     needs: format
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
safl, line 92: Why this?

safl (commented):
A couple of inline questions

.github/workflows/verify_and_publish.yml
mbrsamsung: Github has updated the ubuntu-latest to 24.04 from 22.04 this has broken the build

mbrsamsung: Or at least that was the conclusion when running into that weird bug previous week. Please see [https://github.com/refenv/cijoe/actions/runs/13007722796/job/36278264524](https://github.com/refenv/cijoe/actions/runs/13007722796/job/36278264524) as an example. Or on the main branch not relevant to the existing changes: [https://github.com/refenv/cijoe/actions/runs/13007780475/job/36278443110](https://github.com/refenv/cijoe/actions/runs/13007780475/job/36278443110)

.github/workflows/verify_and_publish.yml
mbrsamsung: This was an debug leftover, sorry for that

.github/workflows/verify_and_publish.yml
safl: Great that you found a work-around! It truly makes no sense that the actions (which is running inside Docker) is affected by the OS of the docker-host. But there must be some preperation stage going on as the GHA runner is started that causes this.... I have added the fix on main, as this broke main as well.
I'll rebase your PRs and give them another look.

.github/workflows/verify_and_publish.yml
```diff
@@ -125,9 +125,16 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
+    - name: Install packages
```
safl, line 128: I don't think you need this anymore as it should be readily available in the docker image.

.github/workflows/verify_and_publish.yml
```diff
@@ -191,11 +200,13 @@ jobs:
       run: |
         make release
 
+
   publish-coverage:
     runs-on: ${{ matrix.os }}
     needs: [examples, unittest]
     container:
       image: ghcr.io/refenv/cijoe-docker:latest
+      options: --privileged
```
safl, line 209: Why is ``--privileged`` required here?

safl (commented):
A couple of comments/questions inline.

.github/workflows/verify_and_publish.yml
safl: Ping @mbrsamsung

.github/workflows/verify_and_publish.yml
```diff
@@ -137,6 +138,8 @@ jobs:
         cijoe --example ${{ matrix.usage_example }}
 
     - name: Run it!
+      env:
+        GHA_RUNNER_TOKEN: "BFGC6XRGGM7PUVLC5NYXARLHQ6FI6"
```
safl, line 142: Why is this token added here?

.github/workflows/verify_and_publish.yml
```diff
@@ -129,6 +129,7 @@ jobs:
     - name: Build and install cijoe from source
       run: |
         pipx uninstall cijoe
+        pipx install build
```
safl, line 132: This should not be needed. ``make deps`` on the next line does pipx-install?

safl (changes requested):
A couple of things inline.

.github/workflows/verify_and_publish.yml
mbrsamsung: There where no other clear way, workflows that runs in a pull request and which are not initiated from a user of the project will not have the secrets included in the run. This will leave the build always failed in that use case.

mbrsamsung: Ah! We are in the other PR. This was meant for GHA example. I guess the splitting up of the commits into PR's did result in this being included erroneously

safl (approved):
lgtm.

PR #148, opened 2025-01-15
Add GHA as a example run

mbrsamsung:


safl (changes requested):
This should be unblocked now?

PR #153, opened 2025-01-30
fiddling...

safl:
testing whether it is possible to update the setup-python action instead of dropping ubuntu-latest...


safl:
nope.. did not do the trick.

Issue #138, opened 2024-12-16
Add support for running packaged scripts without a workflow


safl:
It is currently possible to execute scripts directly, e.g.

```python
cijoe myscript.py --config ...
```

It would be great, to extend this with executing packaged scripts, without a workflow, e.g.:

```python
cijoe qemu.build --config ...
```

In addition, a refactor of the ``--resources`` might be useful, e.g. making it easier to read the list of scripts.

NaddiNadja:
Fixed with #154

PR #154, opened 2025-02-06
feat(cli): add support for running packaged scripts

NaddiNadja:
Packaged scripts that are found by the cijoe resource collection can now be run with command `cijoe <script_name> --config ...`. They can be run in combination with local python files, e.g. `cijoe core.repository_prep foo/bar.py --config ...`

Solves #138


src/cijoe/cli/cli.py
```diff
@@ -367,13 +367,23 @@ def cli_workflow(args):
...
+        else:
+            return step
+
+    steps = list(map(normalise, steps))
+    if len(set(steps)) != len(steps):
         log.error("Duplicate script file names not allowed.")
         sys.exit(1)
 
-    resources = get_resources([step.parent for step in paths])
+    resources = get_resources(list(parent_dirs))
```
safl, line 386: When the user invokes a command like cijoe core.repository_prep foo/bar.py --config .., the order of execution is significant. For example, a repository may need to be prepared before running a build, install, or deploy script. However, using set(), which is unordered, may unintentionally reorder the list of scripts to run.

safl (commented):
One concern inline.

src/cijoe/cli/cli.py
NaddiNadja: The set is only used to add the local scripts' parent directories to the search paths of the resource collector. The order of the steps in remained in the `steps` list :)

safl: Ohhh, awesome then!

safl (approved):
Excellent!

PR #155, opened 2025-02-06
fix(fio): cleanup of fio package and documentation

NaddiNadja:
Remove the `fio_fancy` function and the helper functions associated with it. Update the documentation to actually include links to the fio scripts. Additionally, fix the wrong doc-string for the fio-build script.


safl (approved):
Thanks @NaddiNadja , this is great!

Issue #156, opened 2025-02-10
Feature Request: Improve Built-In Script Parametrization and Input Validation


safl:
With the introduction of https://github.com/refenv/cijoe/pull/154, it is now possible to execute built-in scripts using their "packaged-name" (e.g., ``cijoe qemu.guest_start``). However, not all scripts are currently suitable for direct execution due to the following issues:  

1. **Reliance on Workflow Input Parameters**:  
   Some scripts depend on parameters provided via the workflow ``with-args``  
   instead of retrieving them via configuration files or environment variables.  

2. **Input Validation Issues**:  
   Scripts often fail to handle missing or invalid configurations gracefully,  
   resulting in unexpected behavior.  

Request  
-------

To improve the usability and robustness of these scripts, a review and  
update of their parameterization is needed. Key goals include:  

- Allowing scripts to retrieve input parameters via configuration files or  
  environment variables for seamless execution.  
- Adding proper input validation to ensure graceful handling of missing  
  or invalid configuration entries.  

Why This Is Useful  
-------------------

Enhancing script usability as built-in commands -- especially for the scripts in ``qemu.*``, would make them more convenient and powerful to use.

NaddiNadja:
This has been solved with #163 and #165 :)

PR #157, opened 2025-02-10
Error handling in cijoe scripts

NaddiNadja:
Solves #156

### DIFF
--- a/src/cijoe/core/scripts/cmdrunner.py
+++ b/src/cijoe/core/scripts/cmdrunner.py
@@ -10,6 +10,7 @@ Retargetable: True
 """
 
 import errno
+import logging as log
 
 
 def main(args, cijoe, step):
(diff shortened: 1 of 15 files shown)


safl (commented):
First off this is excellent. One thing that I consider when looking at this is that it might be preferable to use a different naming convention for "script specific inputs".
E.g. how about re-using the naming convention something like:

- ``cijoe.core.cmdrunner``
- ``cijoe.qemu.guest_initialize``

Or with an explicit intention:

- ``cijoe.inputs.core.cmdrunner``
- ``cijoe.inputs.qemu.guest_initialize``

Or,

- ``cijoe.script_input.core.cmdrunner``
- ``cijoe.script_input.qemu.guest_initialize``

Or what do you think?

NaddiNadja:
I have dropped the last commit, as we have decided to refactor the script arguments, and therefore the fallback for step.with is not necessary anymore (which outdates the latest comment as well, as this is no longer part of this PR).

safl (approved):
Excellent!

Issue #124, opened 2024-12-12
Improve the fio package


safl:
- [x] Add the cijoe fio package to package testing / verfication
  - That is, include it in the ``cijoe_packages`` github action workflow
  - See: https://github.com/refenv/cijoe/pull/125
- [ ] Add example (config and workflow) showing how to run / use the fio wrapper
  - Consider simplifying it dramatically, it is currently very convoluted
  - Note: it must exercise the different ways that fio engines are handled e.g. builtin / external / etc.
- [x] Update documentation for the fio package

NaddiNadja:
@safl as far as I can see, there are example workflows and configs for the fio package, added the same day as this issue was opened. Are these okay now that the fio package has been trimmed down or do we need to modify them?

Additionally, does the documentation need to cover more?

safl:
> [@safl](https://github.com/safl) as far as I can see, there are example workflows and configs for the fio package, added the same day as this issue was opened. Are these okay now that the fio package has been trimmed down or do we need to modify them?
> 
> Additionally, does the documentation need to cover more?

If the example workflows and configs cover all scripts and modules, they are sufficient. Otherwise, additional content is needed. In general, coverage should include all scripts and modules, making the setup self-contained.

NaddiNadja:
Closing this, as we have decided to remove the fio package with commit [b3f4eff](https://github.com/refenv/cijoe/commit/b3f4eff75f8c891c113930d7638d9e33494ad4c2) :)

PR #158, opened 2025-02-10
refactor(fio): remove fio wrapper

NaddiNadja:
Removing the fio wrapper, as it is not necessary anymore. This will solve point 2 in #124


safl (approved):
LGTM!

PR #163, opened 2025-02-13
 feat(cli): add subparsers for workflows and scripts 

NaddiNadja:
This commit changes the way cijoe scripts and workflows are run. The interface is now `cijoe {workflow path|script name|script path} [...options]`.

If the positional arg is referencing a stand-alone script, additional cli args can be defined by adding a function to the cijoe script file of the form `add_args(parser: ArgumentParser)`. This will allow us to remove the `step` argument in the cijoe script main function at a later point.

If positional arg is referencing a workflow file, additional positional arguments will be treated as step arguments, defining which steps in the workflow will be run.

### DIFF
--- a/docs/source/200_quickstart.cmd
+++ b/docs/source/200_quickstart.cmd
@@ -8,6 +8,6 @@ cijoe --example
 cijoe --example core.default
 
 # Execute the workflow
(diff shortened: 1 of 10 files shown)


safl (approved):
awesome

PR #165, opened 2025-02-14
Removal of step parameter from cijoe script entry function

NaddiNadja:
⚠️ This is dependent on #163 (so first 2 commits are not part of this PR)

### DIFF
--- a/src/cijoe/core/scripts/example_script_default.py
+++ b/src/cijoe/core/scripts/example_script_default.py
@@ -42,21 +42,21 @@ from cijoe.core.command import Cijoe
 def add_args(parser: ArgumentParser):
     """Optional function for defining command-line arguments for this script"""
     parser.add_argument(
-        "--message", type=str, default=None, help="The message to be printed"
+        "--repeat",
+        type=int,
+        default=1,
+        help="Amount of times the message will be repeated",
     )
 
 
-def main(args: Namespace, cijoe: Cijoe, step: dict):
+def main(args: Namespace, cijoe: Cijoe):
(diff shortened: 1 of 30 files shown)


safl:
@NaddiNadja looks like it is ready once the merge conflict is resolved?

src/cijoe/core/scripts/example_script_default.py
```diff
@@ -42,21 +42,21 @@
...
+def main(args: Namespace, cijoe: Cijoe):
     """Entry-point of the cijoe-script"""
 
-    # Grab message from command-line arguments, if none is given, grab it from
-    # the configuration-file
-    message = getattr(args, "message", None) or cijoe.getconf(
-        "example.message", "Hello World!"
-    )
+    # Grab message from the configuration-file
+    message = cijoe.getconf("example.message", "Hello World!")
```
safl, line 56: Why drop the message from cli-arg / workflow-arg?

safl (commented):
Excellent improvements! One minor question inline.

src/cijoe/core/scripts/example_script_default.py
NaddiNadja: For simplificity's sake, I've changed the `message` variable back to only being defined in the config file. In the last PR, I added it as cli-arg to give an example of how to do this, but I think it's cleaner to just use `repeat` for this cause. This way there are two variables, `message` and `repeat`, and one is an example of how to get config-variables, and the other is an example of how to get cli-arguments. :)

PR #167, opened 2025-02-20
Example and test for `core.testrunner`

NaddiNadja:


NaddiNadja:
The tests pass, but they are not doing what I want to test, so this is not ready yet. Wait for #169

NaddiNadja:
@safl note that I use `cijoe.run_local()` for the tests. This is because when running `cijoe.run()` with `run_local=true` as argument for the testrunner, the call to `cijoe.run()` will be executed on the target machine. So, the testrunner first connects to the target machine defined in the config file at `cijoe.transport.xyz`, then the `Cijoe` object on the target machine will try to connect to the transport defined at `cijoe.transport.xyz` again, which it can't because it doesn't exist as a transport from within the target machine. 

I don't know if there is ever a situation where you would want the initiator to run the testrunner on a target machine, and that the target machine should then run the tests on another target machine. If there is no usecase for `cijoe.run()` when `run_local=false`, I think there might be an idea in always evaluating `cijoe.run()` as `cijoe.run_local()` if it is a test executed by the testrunner with `run_local=false`.

NaddiNadja:
@safl ah yes, I've added a commit that fixes this and changed back to using `cijoe.run()` :)

PR #168, opened 2025-02-24
feat(test): add gha as a example run

NaddiNadja:
It seems like the github action runner installation script has changed substantially from previous.

1. The svc.sh script is not present
2. It is not allowed to run as root or sudo

For this reason the script has been adapted.

Note: I had to create a new pull request to be able to make changes to the branch

### DIFF
--- a/.github/workflows/verify_and_publish.yml
+++ b/.github/workflows/verify_and_publish.yml
@@ -50,6 +50,7 @@ jobs:
 
   unittest:
     needs: format
+    if: false # NADDI
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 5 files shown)


safl:
By default, GitHub runs jobs as root inside a Docker container. As a result, ``config.sh`` didn’t produce ``svc.sh``, and it failed due to root execution.

To resolve this, spin up a guest and run these tasks as a non-root user with password-less sudo. No other changes should be needed.

This setup mimics a typical self-hosted environment, such as a VM or bare-metal host.

safl:
Ok great.

Yes, when trying to run the those scripts inside the docker container, then it is not running in an expected environment.

To get closer to such a thing, then you can spin up a guest wirh qemu which will look an feel more like what is expected.

Come to think of it. It might make sense that cijoe provides such a thing, at least to consome gor its own CI.

It would be great to add a build of .qcow2 image similar to how cijoe is provided as a docker-image, we could provide a .qcow2 image.

We could "dog-food" it all the way and use the scripts from "system-imaging" to do so.

NaddiNadja:
I've closed this, as we have decided to discontinue the gha package with #170

PR #169, opened 2025-02-24
Boolean script arguments

NaddiNadja:
A fix for the boolean script arguments that as of now are incorrectly evaluated with the `bool()` function, as both "true" and "false" are evaluated to `True`.

Also commit "fix(core/cli) ......" fixes an error in the way script arguments were added in the workflow, because this approach broke when running the scripts individually with the cli tool


safl (approved):
lgtm

PR #172, opened 2025-03-03
feat(docs/packages): add cli arguments to the emit script

NaddiNadja:
By adding the help text from running `cijoe <script> --help` to the documentation, we do not need to include the script arguments directly in the in-code documentation.

An example of how it looks:
image


safl (approved):
Excellent

Issue #178, opened 2025-03-10
Unhandled cli-argument (``-l``)


safl:
The log-level logic is somewhat br0k3n. Using this:

```
"""
Virtual block-device using zram
===============================

log-level modifier

"""

import logging as log
from argparse import ArgumentParser, Namespace
from cijoe.core.command import Cijoe

def main(args: Namespace, cijoe: Cijoe):
    """Entry-point of the cijoe-script"""

    log.critical("Foo")
    log.error("THis sis an error")
    log.info("Foo")
    log.warning("Warning")
    log.debug("Foo")

    return 0


if __name__ == "__main__":
    main()
```

Then passing ``-l`` does not increase the log-level. ``-l`` should give +info, and ``-ll`` should give +debug. Yet, this does not happen.

PR #179, opened 2025-03-13
fix(cli/loglevel): add force param to `log.basicConfig()`

NaddiNadja:
According to the docs, the `log.basicConfig()` method does nothing if the root logger already has handlers configured. Apparently this happens before the method is called in the code, and therefore we need the `force` parameter to ensure the configuration is applied.
https://docs.python.org/3/library/logging.html#logging.basicConfig

Additionally, the call to `basicConfig()` is moved to just after the `parse_args()` call to ensure the configuration is applied for the whole run.

Solves #178


NaddiNadja:
@safl I think this might have gone wrong already back when we added that you can run scripts independently from workflows. I added a commit with tests for the log level, to ensure that something like this doesn't happen again. Right now, warning-level logs are first shown with the `-l` parameter - is this still right?

safl:
> @safl I think this might have gone wrong already back when we added that you can run scripts independently from workflows. I added a commit with tests for the log level, to ensure that something like this doesn't happen again. Right now, warning-level logs are first shown with the `-l` parameter - is this still right?

yeah, warnings + info with `-l` then debug etc. with `-ll`.

src/cijoe/cli/cli.py
```diff
@@ -637,18 +637,20 @@ def main(args=None):
         if err:
             log.error("Couldn't parse args; exiting")
             return err
+
```
safl, line 640: This is now also under ``if args is None``,  is that intentional?

tests/core/test_loglevel.py
```diff
@@ -0,0 +1,72 @@
...
+SCRIPT_CONTENT = """
+import logging as log
+def main(args, cijoe):
+    log.critical("critical")
+    log.error("error")
+    log.warning("warning")
+    log.info("info")
+    log.debug("debug")
+    return 0
+"""
```
safl, line 12: The tests here will probably only trigger the case of executing a script-directly, not when executing a workflow. Right? Perhaps that should covered as well?

safl (commented):
A couple of questions.

src/cijoe/cli/cli.py
NaddiNadja: `args` will always be `None`, when initially running `cijoe` from the command line, so this should always be set. This ensures that the log level is not overwritten by anything that might happen in the call to `create_adhoc_workflow()`, which I would say is the intended behaviour.

safl: Right, makes sense.
How about moving the log setup into ``parse_args()``? This ensures that any caller invoking ``parse_args()`` also gets the log level set according to the arguments.

NaddiNadja: Good idea, I've fixed that now

tests/core/test_loglevel.py
NaddiNadja: With log level being set in `parse_args()`, I don't think there would be a difference in calling **cijoe** with a script or with a workflow.

safl (approved):
great!

PR #180, opened 2025-03-13
fix(qemu/wrapper): add termination logic to qemu wrapper and script 

### LINKED ISSUE #82: Add qemu-helper for waiting on a clean guest shutdown
Basically doing the equivalent of this:

```bash
for i in $(seq 1 60); do test -f "$HOME/guests/fedora_40_x86_64/guest.pid" || break; sleep 1; done
```

That is, waiting at most ``timeout`` seconds, in this case **60**, checking whether the guest pidfile exists. This should implement in:

- [ ] qemu-wrapper
- [ ] script using the qemu-wrapper, for re-use in workflows

E.g. a workflow creates a qemu guest, runs cloud-init and shuts the guest down in order to transfer the disk-image of the boot-drive. When doing so, one wants to run e.g. qemu-img convert to compress or change format before transfer. In case the guest is not done shutting the system down, then disk image / guest-file-system corruptions will occur.

NaddiNadja:
Move the SIGTERM signal logic from the `Guest.kill()` to its own method `Guest.wait_for_termination(timeout)`, which allows us to create a script that gives access to this functionality without force-killing the process

This should be what is needed to fix #82

### DIFF
--- a/src/cijoe/qemu/scripts/guest_wait_for_termination.py
+++ b/src/cijoe/qemu/scripts/guest_wait_for_termination.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+"""
+Wait for qemu guest termination
+===============================
+
+Note: The script will not fail if the guest does not exist.
+Note: This script does not itself terminate the qemu guest.
+
+Retargetable: False
+-------------------
+"""
+import logging as log
+import time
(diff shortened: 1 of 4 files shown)


safl (approved):
lgtm!

PR #182, opened 2025-03-13
feat(tests/cli): add tests for `search_for_file()`

NaddiNadja:
I know this is very minimal, but I think it's the last thing that makes sense to add before closing #128


safl (approved):
Great!

PR #184, opened 2025-03-27
fix(cli/configs): add custom append action for config arg

NaddiNadja:
Without the custom config append action, the config cli argument was overwritten to the default value (empty) by the subparser, if the config argument was positioned before the workflow/script argument.


safl (approved):
Nice catch!

Issue #190, opened 2025-06-30
freebsd `nuage-init` vs `cloud-init`


glimchb:
I'm running this freebsd workflow https://github.com/spdk/spdk-ci/pull/69/commits/c01193267de0ec5541e0bca24ae9064df32dfc2a

in this code, `daemonize=False` :

https://github.com/refenv/cijoe/blob/0fdb34fcf51e35aa4a64b4c825c232bf74b829bd/src/cijoe/system_imaging/scripts/diskimage_from_cloudimage.py#L138-L141

but for some reason qemu stays up and waiting for user input and not shutting down

Ctrl-a-x shuts down qemu and then cijoe workflow continues...

Looks like this shutdown is not working?
https://github.com/refenv/cijoe/blob/0fdb34fcf51e35aa4a64b4c825c232bf74b829bd/src/cijoe/system_imaging/auxiliary/cloudinit-freebsd-userdata.user#L20-L24

any insight what is happening ? @safl @NaddiNadja ?

logs are here: https://github.com/glimchb/spdk-ci/actions/runs/15973973830

reading some more I suspect:

```
So it seems that FreeBSD comes with nuageinit, which has a very limited set of supported cloud init formats. 
```

- https://forums.freebsd.org/threads/understanding-the-basic-cloudinit-images.95428/
- https://forum.proxmox.com/threads/setup-cloud-init-freebsd-14-1basic-cloudinit-network-troubleshooting.150670/\
- https://github.com/lima-vm/lima/issues/1508

glimchb:
thank you @NaddiNadja

NaddiNadja:
@glimchb we have decided to not support nuage-init, but we have hosted the FreeBSD 14.2 image with cloud-init installed here, which you can use in your workflow :)
https://refenv.hel1.your-objectstorage.com/cloudinit/freebsd_14.2_x86_64.qcow

glimchb:
got it @NaddiNadja , thanks for the update

any chance you can update documentation and examples with this ? 

We used 
```
cloud.url = "https://object-storage.public.mtl1.vexxhost.net/swift/v1/1dbafeefbd4f4c80864414a441e72dd2/bsd-cloud-image.org/images/freebsd/14.2/2024-12-08/zfs/freebsd-14.2-zfs-2024-12-08.qcow2"
```
in our workflow...

PR #191, opened 2025-08-07
add example for BSD images to system-imaging module

NaddiNadja:
nuage-init is a subset of cloud-init, which is used on the FreeBSD images, so it can't do all the things that cloud-init does, which leads to problems with it not terminating the qemu instance when expected.

We have added the bsd-images repo https://github.com/refenv/bsd-images, where we pre-install cloud-init to BSD images with nuage-init. This PR adds a guest using such an image to the example config files.

Should solve #190


NaddiNadja:
@safl the FreeBSD image is very large, and this PR fails with "System.IO.IOException: No space left on device" from the Github Runner .... Should we just not test the FreeBSD image, or can we increase the memory size?

safl (approved):
lgtm!

PR #192, opened 2025-08-29
Fixes for smaller issues

NaddiNadja:
Fixes for:
- Multiple configs overwriting eachother: merge dicts recursively
- UNIX assumption in Guest.wrapper: allow definition of boot-driver

Added:
- Script for waiting for guest SSH to be ready, should solve #194 
- "copy" button in report which lets you copy the commands to the clipboard, so you don't have to mark it yourself.


safl (approved):
This is great!l functionality and a very nice fix on cloud-init. Lgtm!

PR #195, opened 2025-10-01
refactor(scripts): move "wait fo ssh" script to core package

NaddiNadja:
This script can be useful for other transports than just QEMU guests.


safl:
Excellent, thanks! Pulling this in despite CI failure as that is due to https://www.githubstatus.com/incidents/071h21gptcp0

PR #200, opened 2025-12-17
Simplification and fix of the argument parser

NaddiNadja:


safl (commented):
Are the CI errors related to this?

safl (commented):
Does this require that one now does not explicitly provide ``--workflow``?
I ask, since there are places where CIJOE is in use where it would like convenient not to have to adjust to that. In other words, could providing ``--workflow`` be optional, or does that break with the cleanup done here?

safl (approved):
Excellent, thanks!

PR #201, opened 2025-12-17
feat(workflow): add support for multiline commands in "run" key

### LINKED ISSUE #199: Add support for multi-line commands in core.cmdrunner
Right now, you can add steps to a workflow for running specific commands in the commandline with
```yaml
- name: configure
  run: |
    cd ~/path/that/can/be/long; echo "hello world" > hello.txt
    cd ~/path/that/can/be/long; cat hello.txt
```

For more complex sequences of commands that must be run within the same Paramiko context, it would be great to support multi-line commands like:
```yaml
- name: configure
  run: |
    cd ~/path/that/can/be/long; \
      echo "hello world" > hello.txt; \
      cat hello.txt
```

NaddiNadja:
Before, it was not possible to create a workflow steps like:

```
- name: do_multiple_things 
  run: | 
    cd /some/long/path; \ 
      echo "hello world" > hello.txt; \ 
      cat hello.txt

- name: have_many_arguments 
  run: | 
    ./run_something.sh \ 
      --arg1 value1 \ 
      --arg2 value2 \ 
      --arg2 value3
```

This could end up forcing users to either have very long lines in the workflow, if many arguments had to be given to a tool, *or* have a `cd` prefix for every command in a list of commands if they all had to be executed in a specific directory.

This commit adds multiline support, which follows the syntax of shell scripts where a backslash (\) continues the command onto the next line.

This will fix #199

### DIFF
--- a/src/cijoe/core/resources.py
+++ b/src/cijoe/core/resources.py
@@ -380,7 +380,18 @@ class Workflow(Resource):
                 )
                 return errors
 
-            step["with"]["commands"] = step["run"].splitlines()
+            # non-empty lines in "run"
+            lines: filter[str] = filter(
+                None, map(lambda s: s.strip(), step["run"].splitlines())
+            )
+            step["with"]["commands"] = []
+            buffer = []
+
+            for line in lines:
+                buffer.append(line)
+                if not line.strip().endswith("\\"):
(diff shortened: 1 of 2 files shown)


safl (approved):
LGTM

Issue #198, opened 2025-10-14
Add "depth" and "single_branch" options for core.repository_prep


karlatec:
Enhancement / feature request:
Please add `depth` and `single_branch` options ([something like in Ansible Git](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/git_module.html)) to `repository_prep` script to reduce the size of checked-out repository. Cloning with at full depth and with all branches is troublesome for large repositories, taking way too much disk space.

For [example](https://github.com/spdk/spdk-ci/pull/114) checking out freebsd source repository takes up ~5GB space vs checking out a single branch with depth=1, which takes ~2GB.

NaddiNadja:
Hi! Sorry for not seeing this before now - I've opened a pull request to get this feature :)
I've added it as both an argument in the configuration file, so that you can adjust these args per repository, or in the workflow file, so as to let you just adjust it for every repository. The configuration file will overwrite the setting in the workflow.

In the configuration file:
```toml
[cijoe.repository]
remote = "https://github.com/refenv/cijoe/"
path = "{{ local.env.HOME }}/git/cijoe"
depth = 2
single_branch = true
```

In the workflow:
```yaml
- name: repos
  uses: core.repository_prep
  with:
    depth: 2
    single_branch: true
```

karlatec:
Hey! Sorry for no reply. I saw this last month and planned to get back to it after the holidays and forgot to do so, of course.

Looks great, thank you for that. I see that the commit was already merged, so let me close the issue.

PR #202, opened 2025-12-18
feat(core/scripts): add git args to repository_prep script

NaddiNadja:
This solves issue #198 

I've added it as both an argument in the configuration file, so that you can adjust these args per repository, or in the workflow file, so as to let you just adjust it for every repository. The configuration file will overwrite the setting in the workflow.

In the configuration file:
```toml
[cijoe.repository]
remote = "https://github.com/refenv/cijoe/"
path = "{{ local.env.HOME }}/git/cijoe"
depth = 2
single_branch = true
```

In the workflow:
```yaml
- name: repos
  uses: core.repository_prep
  with:
    depth: 2
    single_branch: true
```

### DIFF
--- a/src/cijoe/core/scripts/repository_prep.py
+++ b/src/cijoe/core/scripts/repository_prep.py
@@ -32,7 +32,28 @@ Retargetable: True
 
 import errno
 import logging as log
+from argparse import ArgumentParser, _StoreAction
 from pathlib import Path
+
+
+def add_args(parser: ArgumentParser):
+    class StringToBoolAction(_StoreAction):
+        def __call__(self, parser, namespace, values, option_string=None):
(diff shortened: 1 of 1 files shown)


safl (approved):
Lgtm

src/cijoe/core/scripts/repository_prep.py
```diff
@@ -56,13 +82,14 @@ def main(args, cijoe):
             return err
 
         err, _ = run(
-            f"[ ! -d {repos['path']} ] &&"
-            f" git clone {repos['remote']} {repos['path']} --recursive"
```
glimchb, line 60: maybe `--recursive` should also become an argument...

PR #205, opened 2026-05-05
Rename workflow to task

safl:
The YAML intentionally mirrors GitHub Actions in the small (`name`, `uses`, `with`, `run`); that part is on purpose and stays. But calling the file itself a "workflow" oversells the resemblance: there are no matrices, no branches, no conditionals, no expressions. A step is either a plain list of shell commands (`run:`) or an invocation of an expressive Python cijoe script (`uses:` + `with:`); that's the whole vocabulary. The shared name sets the wrong expectation about what cijoe does. "Task" keeps the familiar step keys while signalling that the surrounding orchestration is deliberately minimal, and that the expressive power lives in the Python scripts, not in the YAML.

This is also part of the runway toward a v1.0 release; concept-level naming is the kind of thing that has to be settled before stabilizing the API, not after.

## Backwards compatibility

The rename is staged with deprecation aliases on every user-facing surface (filename, env var, config key, CLI flag, resource subdir, Python class) so existing setups keep working with a warning rather than a break.


naddinadja (approved):
Looks good :)

PR #206, opened 2026-06-28
feat(system_imaging): stage disk images from oras:// artifacts

safl:
cijoe's system-imaging can build a disk image from a cloud image and download a
prebuilt one from a URL, but it cannot consume an image published as an oras://
artifact (an OCI registry blob), which is how the nosi guest-images are shipped.
Such artifacts have no stable download URL, so consumers have each reimplemented
the bearer-token resolution to fetch them.

This adds a diskimage_from_oras script to the system-imaging plugin. For each
system-imaging.images.<name> that declares an oras source, it resolves the
reference with withcache.oras, pulls the blob with curl (retrying and resuming,
which matters against ghcr throttling), decompresses it, and converts it to the
qcow2 at disk.path. The pull is skipped when the pinned image is already staged,
keyed on the content digest. withcache becomes a dependency, so any environment
that has cijoe (including the nosi container) has it too.

Consumers can then drop their own staging scripts and point a
system_imaging.diskimage_from_oras step at the oras reference. Validated with
black, isort, ruff, and mypy, plus unit tests for the digest-keyed skip and the
image selection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ Sections with "(Upcoming)" describe changes on the roadmap for CIJOE.
 Changes on the `master` branch, from the latest version tag up to and including
 HEAD can be subject to a git rebase.
 
+## 0.9.59 (Upcoming)
+
+Added oras:// support to system-imaging via the diskimage_from_oras script: it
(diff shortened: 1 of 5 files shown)


naddinadja (approved):
lgtm! :)

PR #207, opened 2026-06-29
Release v0.9.60: Node 24 CI, low-risk tech debt, version bump

safl:
## Summary

This PR rolls four small chores into a v0.9.60 release.

The CI work bumps every action in `verify_and_publish.yml` to a release whose `action.yml` declares `node24`, so GitHub stops forcing them onto Node 24 with a deprecation notice on every run. The pins are `actions/checkout` to v5.0.1, `actions/setup-python` to v6.0.0, `actions/upload-artifact` to v6.0.0 and `actions/download-artifact` to v7.0.0. Earlier versions in the v5/v6 line of the artifact actions still ship Node 20, so they were ruled out. `actions/upload-artifact` flipped the default of `include-hidden-files` to `false` in v4.4, which silently drops dotfile paths like `.coverage`; the three steps that upload `.coverage` get an explicit `include-hidden-files: true`. On macOS runners the image now pre-taps `aws/tap`, which Homebrew refuses to load without explicit trust and prints a multi-line warning on every `brew update`; cijoe pulls nothing from `aws/tap`, so each `brew update` is preceded by `brew untap aws/tap || true`.

The tech-debt pass drops 22 unused imports flagged by ruff F401 across `docs/`, `src/`, and `tests/`, fixes an unused `state` destructuring at `src/cijoe/core/scripts/cmdrunner.py:65`, and removes three lines of commented-out dead code in `src/cijoe/linux/null_blk.py`.

The post-rename cleanup resolves three leftover references to "workflow" after the recent rename to "task": the Sphinx sidebar pattern in `docs/source/conf.py` excluded a directory that no longer exists, the example task header had a "GitHUB" casing typo, and the bash completion's default filename was the legacy `cijoe-workflow.yaml` so step-name completion silently failed once users moved to `cijoe-task.yaml`. It now prefers the new name and falls back to the legacy one for users mid-migration.

Closes with a version bump to v0.9.60 via the same two-file pattern as prior bumps (`src/cijoe/core/__init__.py` and `.github/workflows/cijoe_docker.yml`).

## Test plan

Watch `verify_and_publish` go green on this PR, including the macOS legs of `unittest` where the brew warning originated. Confirm the `publish-coverage` job finds `.coverage` artifacts to combine, and inspect a macOS unittest log to confirm the aws/tap warning and Node 20 deprecation note are gone.


naddinadja (approved):
lots of good, small fixes! :)

PR #208, opened 2026-06-29
chore: low-risk technical debt pass

safl:
## Summary

Two scoped cleanup commits.

The first drops 22 unused imports flagged by ruff F401 across `docs/`, `src/`, and `tests/`. None of them were API re-exports, confirmed by grepping for each name across the tree before applying `ruff --fix`. While in the same area, it also fixes an unused `state` destructuring at `src/cijoe/core/scripts/cmdrunner.py:65` (a long-standing pyright warning) and removes a three-line commented-out `rmdir` in `src/cijoe/linux/null_blk.py` whose author note had marked it "not useful yet".

The second resolves three leftover references to "workflow" after the recent rename to "task". The Sphinx sidebar pattern in `docs/source/conf.py` still excluded `workflows**` against a directory that has since become `tasks/`. The example task's header had a "GitHUB" casing typo. The bash completion at `src/cijoe/core/auxiliary/cijoe-completions` only knew the legacy default filename `cijoe-workflow.yaml`, so step-name completion silently failed once a user moved to `cijoe-task.yaml`; it now prefers the new name and falls back to the legacy one for users mid-migration.

Items the survey flagged but were intentionally left out: `src/cijoe/core/analyser.py` (no in-tree callers but possibly external API), the open TODOs in `null_blk.py`, `cli.py`, `testrunner.py`, and the report template, the `tomli`/`tomllib` conditional in `resources.py` (3.10 has neither, so the conditional stays), and every deliberate back-compat shim from the rename.

## Test plan

Confirm `verify_and_publish` is green on this branch and that no test depends on any of the removed imports.

### DIFF
--- a/docs/emit.py
+++ b/docs/emit.py
@@ -6,7 +6,6 @@ import os
 import re
 import subprocess
 from pathlib import Path
-from pprint import pprint
 
 from jinja2 import Template
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -45,7 +45,7 @@ html_sidebars = {
     "scripts**": [],
     "testrunner**": [],
     "usage**": [],
-    "workflows**": [],
+    "tasks**": [],
 }
 html_theme_options = {
     "header_links_before_dropdown": 8,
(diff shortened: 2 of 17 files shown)


safl:
Folded into #207 with the v0.9.60 version bump.

PR #209, opened 2026-06-29
chore(configs): bump example distros to Debian 13 and Fedora 44

safl:
## Summary

Rolls the example configs and the `test_config_images` CI matrix forward to Debian 13 (trixie) and Fedora 44, both currently the latest stable from their respective vendors. Ubuntu 24.04 LTS and FreeBSD 14 are left as-is; Ubuntu noble is still the active LTS and the FreeBSD URL is a credentialed mirror that needs separate attention.

Mechanical replacements across 12 files: `debian-12-{x86_64,aarch64}` -> `debian-13-*` (including the `bookworm` -> `trixie` URL path and `debian-12-generic-amd64-daily.qcow2` filename change) and `fedora-43-x86_64` -> `fedora-44-x86_64` (including the cloud-image filename bump from `Fedora-Cloud-Base-Generic-43-1.6` to `Fedora-Cloud-Base-Generic-44-1.7`, the patch level published with the new release). Cloud-image URLs verified via curl HEAD; both return 200.

One judgement call: `src/cijoe/core/configs/example_config_get_put.toml` dropped its `disk.url`/`disk.url_checksum` shortcut for `debian-13-aarch64`. refenv only hosts a `debian-bookworm-arm64` mirror under `refenv.fra1.digitaloceanspaces.com`, with no trixie variant available, and `disk.url` is only a cache shortcut on top of `cloud.url` anyway. The other `disk.url` entries point at `cijoe-system-imaging.s3...backblazeb2.com` paths that already 404 for both debian-12 and debian-13 (pre-existing dead URLs), so they were left in the diff for shape consistency; the workflow falls back to `cloud.url` regardless.

## Test plan

Watch `test_config_images (debian-13, x86_64)`, `test_config_images (debian-13, aarch64)` and `test_config_images (fedora-44, x86_64)` go green on this PR. Confirm the `examples` jobs that exercise QEMU guests (`core.get_put`, `core.testrunner`, `qemu.guest_*`, `linux.null_blk`) still pass with the new image names. Re-run if the cloud.debian.org mirror flakes mid-download (the same flake that hit the v0.9.60 tag run is unrelated to the version bump).

### DIFF
--- a/.github/workflows/verify_and_publish.yml
+++ b/.github/workflows/verify_and_publish.yml
@@ -206,14 +206,14 @@ jobs:
       fail-fast: false
       matrix:
         image:
-        - debian-12
-        - fedora-43
+        - debian-13
+        - fedora-44
         - ubuntu-2404
         - freebsd-14
         architecture:
         - x86_64
         include:
-        - image: debian-12
+        - image: debian-13
           architecture: aarch64
 
     steps:
--- a/src/cijoe/core/configs/example_config_get_put.toml
+++ b/src/cijoe/core/configs/example_config_get_put.toml
@@ -40,7 +40,7 @@ system_args.raw = """\
 #
 # You can uncomment it in your config here, or provide as argument as a task step-argument
 #
-#initialize.diskimage = "debian-12-aarch64"
+#initialize.diskimage = "debian-13-aarch64"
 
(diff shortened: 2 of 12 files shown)


naddinadja (approved):
lgtm :)

PR #210, opened 2026-08-04
fix(docs/config): use correct format for env config

naddinadja:
The format of the example code was not valid TOML, confusing users. Also, I added an "optionally" to the sentence to clarify that this key is not necessary to make cijoe run.


safl (approved):
lgtm

